PR #4, opened 2026-05-12
fix: capture Codex event transcript rows

### LINKED ISSUE #3: fix: capture clean Codex transcript event rows
## Problem

Codex transcripts contain `response_item` message rows that can include injected context such as AGENTS/system/developer instructions. If ambient capture treats those rows as conversation turns, `iai-mcp` can store prompt context instead of the actual user/assistant exchange.

Codex also emits cleaner conversational rows as `event_msg` entries:

- `payload.type=user_message`
- `payload.type=agent_message`

Those are the safer source for Codex transcript capture.

## Expected fix

For Codex transcripts, capture user/assistant turns from `event_msg` rows and skip `response_item` message rows. Keep existing Claude Code transcript parsing unchanged, and share the parser between inline capture and no-spawn deferred capture so both paths behave the same.

dergachoff:
## Problem

Codex transcripts include `response_item` message rows that can contain injected context such as AGENTS/system/developer instructions. Treating those rows as conversation turns can store prompt context instead of the actual user/assistant exchange.

## Root cause

The transcript parser handled generic message rows but did not distinguish Codex's noisy `response_item` rows from cleaner `event_msg` conversation rows.

## Fix

Adds a shared transcript-turn extractor used by both inline capture and no-spawn deferred capture.

For Codex transcripts, it skips `response_item` rows and captures:

- `event_msg` + `payload.type=user_message` as `user`
- `event_msg` + `payload.type=agent_message` as `assistant`

Existing Claude Code transcript parsing stays unchanged.

## Validation

- `uv run --python 3.12 --with pytest --with pytest-rerunfailures pytest tests/test_capture_transcript_no_spawn_defer.py tests/test_capture_hooks_codex.py -q`
- `uv run --python 3.12 --with 'ruff>=0.5,<0.6' ruff check src/iai_mcp/capture.py tests/test_capture_transcript_no_spawn_defer.py`
- `git diff --check`

Closes #3.

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -35,7 +35,7 @@ import time
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 # R3 deviation [Rule 3 - blocking import cost]: `iai_mcp.embed` pulls
 # in transformers + torch (~2.9s cold import). Loading capture.py for the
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks @dergachoff — merged. Credited you as co-author on the commit for the Codex transcript work. Appreciated.

CodeAbra:
Merged locally as squash with co-author credit.

PR #10, opened 2026-06-17
Support macOS Keychain credentials and make --bare optional

Marsu6996:
### Problem

On macOS with an active Claude subscription, the nightly consolidation never uses the subscription LLM path. Two independent causes:

1. **Keychain credentials.** `verify_credentials_subscription()` only reads `~/.claude/.credentials.json`, but on macOS Claude Code stores OAuth credentials in the **login Keychain** (generic-password item `Claude Code-credentials`). On a normal desktop-app / `claude /login` install the file doesn't exist, so the check returns `credentials_file_missing` even though the subscription is active. `claude /login` does not create the file on macOS.

2. **`--bare` + auth.** `_build_cmd()` hardcodes `--bare`. On recent Claude Code (tested **v2.1.179**), `claude --bare -p …` returns `{"is_error": true, "result": "Not logged in · Please run /login"}` for Keychain-stored credentials (and for a manually-created `.credentials.json`, and for a `CLAUDE_CODE_OAUTH_TOKEN` from `claude setup-token`). The identical command **without** `--bare` authenticates normally. So in this version `--bare` skips the credential path this setup relies on. (Not the Apple-billing issue anthropics/claude-code#46847 — this account is `stripe_subscription`.)

### Changes

- `verify_credentials_subscription()` falls back to reading the macOS Keychain item when `~/.claude/.credentials.json` is absent.
- `--bare` is gated behind a new env var `IAI_MCP_CLAUDE_BARE` (**default on — behavior unchanged**). Affected macOS users can set `IAI_MCP_CLAUDE_BARE=0` to drop the flag.

### Notes / open question for maintainers

- Default behavior is unchanged; existing tests pass (`tests/test_claude_cli.py`, 22 passed — including `assert "--bare" in args`).
- Possible follow-up (not changed here): the `cost_usd > 0` "API billing detected" tripwire reads `data.get("cost_usd")`, but `claude -p --output-format json` emits **`total_cost_usd`** (and reports a non-zero value even on subscription usage). Worth reviewing what signal best distinguishes a paid-API call from subscription use.

### Environment

macOS 26.5.1 (arm64) · Claude Code CLI v2.1.179 · iai-mcp 1.1.1 · Python 3.11.

### DIFF
--- a/src/iai_mcp/claude_cli.py
+++ b/src/iai_mcp/claude_cli.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import os
 import subprocess
+import sys
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


CodeAbra:
Thanks for this — you nailed both the macOS Keychain fallback and the `--bare` auth issue, and the writeup made it easy to verify. Cherry-picked your fix into `main` for v1.1.2 and credited you as co-author on the commit. Closing this out since it has landed there. Really appreciate the catch. 🙏

PR #2, opened 2026-05-11
feat: add Codex capture hook installer support

dergachoff:
## Summary
Adds Codex as an optional `capture-hooks` target while keeping Claude as the default.

The Codex Stop hook reads Codex hook JSON from stdin, captures the transcript with `iai-mcp capture-transcript --no-spawn`, exits 0, and stays silent on stdout when skipping capture. The installer patches `~/.codex/hooks.json`, preserves unrelated hooks, and status/uninstall handle Codex separately.

Closes #1

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor (no behaviour change)
- [x] Documentation
- [ ] Build / tooling

## Affected areas
- [x] Capture path
- [ ] Recall / retrieval
- [ ] Consolidation / sleep cycles
- [ ] Daemon lifecycle
- [ ] Storage / encryption at rest
- [ ] MCP wrapper
- [ ] Bench harness
- [x] CLI / doctor
- [x] Documentation
- [ ] Other: ___

## Testing
- [x] New tests added for changed behaviour
- [x] `uv run --python 3.12 --with pytest --with pytest-rerunfailures pytest tests/test_capture_hooks_codex.py tests/test_capture_transcript_no_spawn_defer.py -q`
- [x] `uv run --python 3.12 --with 'ruff>=0.5,<0.6' ruff check src/iai_mcp/cli.py tests/test_capture_hooks_codex.py`
- [x] `bash -n deploy/hooks/iai-mcp-codex-session-capture.sh`
- [x] `git diff --check`

## Benchmarks
Not run. This is hook installer and documentation work; it does not change retrieval ranking, consolidation, embedding, or capture parsing semantics.

## Notes for reviewers
This PR intentionally limits Codex support to ambient Stop-hook capture. It does not change iai-mcp recall mechanics.

Docs use `[features].hooks = true` only as troubleshooting for disabled or older installs; modern Codex hooks are stable/default-enabled.

Issue #1, opened 2026-05-11
feat: add Codex capture hook installer support


dergachoff:
## Problem
iai-mcp has ambient capture installer support for Claude Code, but not for
Codex CLI. Codex can run lifecycle hooks and MCP stdio servers, so users should
not need to hand-edit `~/.codex/hooks.json` to capture completed Codex sessions.

## Evidence
Current Codex builds support `Stop` hooks loaded from `hooks.json`. Codex
`rust-v0.124.0` made hooks stable, and `rust-v0.129.0` made
`[features].hooks` the canonical feature flag while keeping `codex_hooks` as a
legacy alias.

For iai-mcp, the required integration is a Stop hook that can read the session
transcript path and call `iai-mcp capture-transcript --no-spawn`.

## Expected fix
Add Codex as an opt-in `capture-hooks` target:
- install a Codex Stop hook under `~/.codex/hooks/`
- patch `~/.codex/hooks.json`
- preserve unrelated hooks
- uninstall only the iai-mcp hook entry
- document Codex MCP and hook setup

PR #12, opened 2026-06-18
Windows port: platform-agnostic IPC, fcntl shim, POSIX signal guards, Task Scheduler daemon

danielhertz1999-bit:
## Windows port — complete & validated on a real Windows 11 box

This PR (head = `danielhertz1999-bit:main`) is the **canonical, complete** Windows port. It supersedes #18 (an earlier rework, now ~84 commits behind). Everything below is verified on Windows 11 / Python 3.12 / Rust 1.96 (MSVC) with the project built from source.

### Platform port
- IPC: Unix-domain socket → **TCP loopback** with an ephemeral port persisted to a per-endpoint `.port` file (`_ipc.py`); honors `IAI_DAEMON_SOCKET_PATH` for isolation.
- **Auth-token handshake** on the loopback TCP (32-byte token, ACL-restricted via `icacls`, `compare_digest` validation) — closes the "loopback reachable by any local process" concern from @warplayer's review. Token file is also per-endpoint isolated.
- `fcntl` → `_filelock` shim (`msvcrt`, offset-preserving, POSIX-blocking emulation); `resource`/`signal`/`os.geteuid`/`pwd` guarded; `os.kill(pid,0)`→`psutil.pid_exists`.
- Daemon installer via **Task Scheduler** (`schtasks` XML); **PowerShell hooks**; `%APPDATA%` log paths; `icacls` key security.
- `package`-level `__main__.py` (the hooks call `python -m iai_mcp`); UTF-8 forced on all text I/O + std streams; `os.replace` for atomic moves.
- Rust `Cargo.toml` macOS-only features gated; `pip install -e .` builds the native ext + hnswlib clean.

### Validation
- Builds + imports clean; **core test suite ~90%+** green.
- **Full daemon/socket test suite ported** to the cross-platform transport (with the auth handshake) — passes or cleanly skips POSIX-only cases; no hangs. Includes real-daemon-spawn tests.

### Bugs fixed (surfaced by the port)
`_patch_claude_desktop_config` crash · `_filelock` offset/blocking · `doctor.check_b` socket-file gate · `capture._pid_is_alive` (`os.kill(pid,0)` is `CTRL_C_EVENT` on Windows) · port-file & token-file isolation.

Happy to rebase onto the latest `main` whenever you're ready to merge. Thanks for tracking this!

_🤖 Validated with [Claude Code](https://claude.com/claude-code) on a real Windows 11 build_


warplayer:
Hi @danielhertz1999-bit 👋 — I'm not a maintainer, just someone who came across this PR and happened to have a Windows 11 machine, so I took the branch for a spin to give you some independent test data. Really nice work — the platform-abstraction approach (the `_ipc` / `_filelock` shims plus explicit per-OS branches) is clean and easy to follow. Everything below is offered as a fellow contributor; please treat it as FYI and ignore anything you've already got in flight.

**Environment:** Windows 11 Home (build 26200), fresh `gh pr checkout 12`, Python 3.12.10 venv, Rust 1.96.0 (`x86_64-pc-windows-msvc`, VS BuildTools MSVC 14.51), Node 24.15.

### What worked great (actually ran it, not just static checks)
- **`pip install -e ".[dev]"` succeeds end-to-end, including the Rust build.** `iai_mcp_native.cp312-win_amd64.pyd` compiles and imports (`embed` + `graph`), `hnswlib` builds from sdist via `cl.exe`, and the whole scientific stack resolves to cp312 wheels (numpy 2.2.6, scipy 1.17.1, numba 0.65.1, pyarrow 19, cryptography 49, keyring via `pywin32-ctypes`). The `--no-default-features` Rust build keeps the macOS `accelerate`/`metal` features out, so no Apple-framework link errors.
- **`mcp-wrapper` builds clean** (`tsc` → `dist/index.js`) on Node 24.
- **`daemon install` actually round-trips through Task Scheduler.** The generated UTF-16 XML is accepted by `schtasks /Create /XML`, registers `\iai-mcp-daemon` with `Task To Run = …\pythonw.exe -m iai_mcp.daemon`, `Start In = %APPDATA%\iai-mcp\logs`, `Run As = <user>`; `daemon uninstall` cleanly `/End`+`/Delete`s it.
- **`_ipc` TCP-loopback works in a live round-trip** — server binds `127.0.0.1:<ephemeral>`, writes `.daemon.port`, client discovers it and echoes, port file removed on shutdown.
- **`capture-hooks install` writes correct Windows wiring** — `powershell -ExecutionPolicy Bypass -File "…ps1"` for Stop/UserPromptSubmit/SessionStart with the right timeouts + matcher; `status` → WIRED; `uninstall` reverts cleanly.

### A few things I ran into
1. **`pytest` can't be collected on Windows** — 5 test modules import POSIX-only modules at top level, so collection aborts before any test runs (`Interrupted: 5 errors during collection`):
   - `tests/test_capture_queue.py`, `tests/test_doctor_lock_probe.py`, `tests/test_live_e2e_gate.py`, `tests/test_lock_starvation.py` → `import fcntl`
   - `tests/test_daemon_fdlimit_and_fsm.py` → `import resource`

   The `src/` tree was nicely ported to the `_filelock`/`_ipc` shims, but these tests still import `fcntl`/`resource` directly. With `--continue-on-collection-errors`, a partial run (~31% of the suite before the 60s daemon socket-bind timeouts led me to cut it at 6m35s) gave **957 passed / 98 failed / 74 skipped / 7 errors**. The large majority of those failures are the **test suite itself assuming POSIX** — fake daemons built on `asyncio.start_unix_server` / `open_unix_connection` / `socket.AF_UNIX`, plus `signal.SIGKILL`, `os.geteuid`, `/tmp` paths, and `0o600` mode-bit assertions — i.e. the same gap as the collection errors: `src/` was ported but the tests weren't. A smaller set look like genuine latent cross-platform issues the port *surfaces* (not regressions in this PR), e.g.: file I/O without `encoding="utf-8"` falling back to cp1252 (`bench/contradiction_longitudinal_claude.py:817` → `UnicodeEncodeError` on `Δ`); Windows file-replace semantics in crypto-key rotation (`WinError 183`); and a couple of hippo lock-escalation failures consistent with the `LOCK_SH` divergence in #3. Happy to share the full categorized list.

2. **`capture-hooks install` crashes if the MCP wrapper isn't built yet *and* Claude Desktop is installed.** `_patch_claude_code_config` handles a missing wrapper by writing a placeholder (`cli/_capture.py:452-470`), but `_patch_claude_desktop_config` calls `_build_iai_mcp_server_entry()` without the same guard (`cli/_capture.py:404,422`), so it raises an uncaught `FileNotFoundError` *after* already writing the hooks + `settings.json`. Repro on my box (Desktop installed, wrapper unbuilt): install writes the hooks, then traceback + non-zero exit. Building `mcp-wrapper` first makes it exit 0. Since the README/handoff flow installs hooks before building the wrapper, it's easy to trip. (Not strictly Windows-specific.)

3. **`_filelock` shared locks aren't actually shared on Windows.** `msvcrt.locking` is exclusive-only, so the shim maps `LOCK_SH` to the same exclusive lock as `LOCK_EX`. Cross-process repro: process A takes `LOCK_SH`, process B's `LOCK_SH | LOCK_NB` returns `EWOULDBLOCK` (on POSIX the second reader gets the lock). That changes the multi-reader path in `hippo/_db.py` (`LOCK_SH` ~lines 253/306), `lock_protocol.acquire_client_shared_nb`, and the `doctor` shared-probe at `doctor/_lifecycle_checks.py:186` (it would read "exclusively held" when only a reader holds it). The errno-normalization itself is correct — it's the lock *mode* that differs. A faithful shared lock on Windows would need `LockFileEx` (via `ctypes`), or this could just be documented as a known divergence.

4. **Minor: `_filelock.flock()` moves the file offset.** The Windows path does `os.lseek(fd, 0, SEEK_SET)` before locking, whereas `fcntl.flock` leaves the offset untouched (observed 7 → 0 after a `flock` call). Probably harmless since the current callers seek explicitly, but it's a behavioral difference. (Also, per your own code comment, blocking `LK_LOCK` raises after ~10s vs POSIX blocking forever — could surface as spurious failures under long contention.)

5. **Design question (not a bug): Windows IPC is a loopback TCP port with no auth token.** The Unix-domain socket's access was bounded by filesystem perms; `127.0.0.1:<port>` is reachable by any local process/user. CONTRIBUTING.md flags "network-exposed surface beyond the existing local UNIX socket" as sensitive, so I figured it was worth surfacing — a Windows named pipe (`\\.\pipe\…`) would preserve ACL-based access control if that property matters here. Could well be a deliberate trade-off.

(FWIW the em-dash-renders-as-`�`-on-a-cp1252-console thing you already flagged in the handoff is indeed cosmetic — the on-disk XML is UTF-16 and `schtasks` reads it fine.)

Happy to share the exact repro scripts, open issues, or send a small PR for the test-collection guards and/or the `_patch_claude_desktop_config` fallback — only if it's useful and doesn't cut across what you're already doing. Thanks for porting this to Windows! 🙏

CodeAbra:
hey guys i would really appreciate some help with windows porting
please feel free to open PRs

dealing with repo rn

danielhertz1999-bit:
Thanks for the warm welcome and for sharing that Windows support is a priority! This PR covers Steps 1–7 of a full Windows port: platform-agnostic IPC (Unix sockets → TCP loopback), `fcntl`/`resource`/`signal`/`os.geteuid` shims, a Task Scheduler daemon installer, and PowerShell hook scripts. Happy to iterate on any feedback.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

danielhertz1999-bit:
@warplayer — this is incredibly thorough, thank you! Really appreciate the live test data. Addressing your points:

**1. Test collection errors (fcntl / resource) — fixed** in commit [269e90a](https://github.com/danielhertz1999-bit/iai-personal-memory-engine/commit/269e90a). The 5 test files now import from the `_filelock` shim instead of `fcntl` directly; the `resource` import is guarded with `sys.platform != "win32"` and the `TestRaiseFdLimitClampsToHard` class is `@skipif`'d on Windows. The remaining failures you described (POSIX-assuming fake daemons, `/tmp` paths, cp1252 encoding, `WinError 183` key rotation) are on the backlog — good to have them categorized.

**2. `_patch_claude_desktop_config` crash when wrapper unbuilt** — confirmed, this is a real bug (not Windows-specific). Adding a `FileNotFoundError` guard to `_build_iai_mcp_server_entry()` at `cli/_capture.py:404,422` is a clear fix; will include that in the next commit.

**3. `_filelock` `LOCK_SH` maps to exclusive on Windows** — correct, this is a known divergence documented in `_filelock.py`. A faithful shared lock via `LockFileEx` (ctypes) would be the right fix for `hippo/_db.py`'s multi-reader path and the doctor shared probe; flagged as a follow-up issue since it's a bigger change.

**4. `flock()` moves file offset on Windows** — good catch. The `lseek` before `msvcrt.locking` was to reset to byte 0 for the lock range; callers do seek explicitly, but aligning behavior with POSIX (leave offset untouched) is the right call. Will fix.

**5. TCP loopback auth / named pipe** — acknowledged trade-off. Windows named pipes (`\.\pipe\iai-mcp`) with ACL-based access would be the proper equivalent. Adding it as a future hardening issue.

Thanks again — this kind of hands-on validation is exactly what's needed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

warplayer:
<html>
<body>
<p>Hey @danielhertz1999-bit 👋 — back with more live Windows data. This time I went past the install <em>mechanics</em> and did a <strong>full end-to-end daemon install</strong> (<code>daemon install</code> → <code>daemon start</code> → <code>daemon status</code>, plus <code>capture-hooks install</code>) on Win11. Good news first: the auth-token handshake from #2 works exactly as designed — correct token round-trips, and a raw <code>connect()</code> with no/wrong token gets dropped immediately, with the token file ACL'd to the user. 👍</p>
<p>The flip side: chasing why the installed daemon never actually came up, I hit <strong>three independent bugs that each block the daemon from starting on Windows</strong>. (Caveat/correction to my earlier review: when I said "daemon install round-trips through Task Scheduler," that was the <em>registration</em> + a separate manual IPC test — I hadn't confirmed the installed daemon process actually starts and serves. It doesn't, until these are fixed.) All three are small fixes:</p>
<h3>1. Crypto key corrupted by a text-mode write → daemon crashes on startup</h3>
<p><code>CryptoKey._try_file_set</code> (<code>src/iai_mcp/crypto.py</code>) opens the key file with <code>os.open(...)</code> and no <code>O_BINARY</code>. On Windows that's <strong>text mode</strong>, so <code>os.write()</code> translates any <code>0x0A</code> byte in the 32-byte AES key to <code>0x0D 0x0A</code>. Result: ~12% of generated keys (<code>1 − (255/256)³² ≈ 11.8%</code>) land on disk as 33 bytes and are then rejected on read:</p>
<pre><code>iai_mcp.crypto.CryptoKeyError: crypto key file at C:\Users\&lt;u&gt;\.iai-mcp\.crypto.key has wrong length 33 (expected 32)
</code></pre>
<p>I hit it on the first install; the on-disk hex showed <code>…2599f3</code> <strong><code>0d0a</code></strong> <code>62288e…</code> — a single <code>0x0A</code> expanded to CRLF. Fix:</p>
<pre><code class="language-python"># crypto.py, _try_file_set
flags = os.O_CREAT | os.O_EXCL | os.O_WRONLY | getattr(os, "O_BINARY", 0)
fd = os.open(str(tmp), flags, 0o600)
</code></pre>
<p><code>getattr(os, "O_BINARY", 0)</code> is <code>0</code> (a no-op) on POSIX. Verified a worst-case all-<code>0x0A</code> key now round-trips at exactly 32 bytes. (This also covers <code>rotate()</code>, which goes through the same writer.)</p>
<h3>2. <code>os.kill(pid, 0)</code> liveness probe → <code>WinError 87</code></h3>
<p><code>lifecycle_lock._is_pid_alive</code> (<code>src/iai_mcp/lifecycle_lock.py:54</code>) uses the POSIX <code>os.kill(pid, 0)</code> idiom. On Windows <code>os.kill</code> only honors <code>CTRL_C_EVENT</code>/<code>CTRL_BREAK_EVENT</code> and otherwise calls <code>TerminateProcess</code>, so signal <code>0</code> raises:</p>
<pre><code>OSError: [WinError 87] The parameter is incorrect
</code></pre>
<p>This fires whenever the daemon finds an existing lock (e.g. after an unclean restart), so stale-lock takeover crashes instead of recovering. (It's also a little unsafe — a different signal value there could terminate the target.) Since the function already uses <code>psutil</code> just below, the fix is to probe via psutil on Windows and leave the POSIX path untouched:</p>
<pre><code class="language-python">if sys.platform == "win32":
    try:
        import psutil
    except ImportError:
        log.debug("lifecycle_lock: psutil unavailable on Windows; assuming pid=%d live", pid)
        return True
    if not psutil.pid_exists(pid):
        return False
else:
    try:
        os.kill(pid, 0)
    except ProcessLookupError:
        return False
    except PermissionError:
        return True
# ...existing psutil cmdline check follows unchanged...
</code></pre>
<p><code>test_lifecycle_lock*</code> → 18/18 pass with this.</p>
<h3>3. Task <code>WorkingDirectory</code> under <code>%APPDATA%</code> → <code>ERROR_DIRECTORY</code>, daemon never launches</h3>
<p><code>_render_schtasks_xml</code> (<code>src/iai_mcp/cli/_daemon.py</code>) sets the scheduled task's <code>&lt;WorkingDirectory&gt;</code> to <code>%APPDATA%\iai-mcp\logs</code>. Task Scheduler's launch context rejects <strong>any</strong> <code>%APPDATA%\Roaming</code> working dir with <code>ERROR_DIRECTORY</code> (Win32 267 / <code>0x8007010B</code>) <strong>even though the directory exists</strong> — so <code>daemon start</code> silently fails (<code>LastTaskResult = 0x8007010B</code>) and nothing ever runs. I isolated it by swapping just the working dir:</p>

WorkingDirectory | LastTaskResult
-- | --
%APPDATA%\iai-mcp\logs | ❌ 0x8007010B
%APPDATA%\iai-mcp | ❌ 0x8007010B
%USERPROFILE% | ✅ 0x00041301 (running)
repo root | ✅ running


<p>The daemon writes its logs via an absolute path and doesn't depend on its CWD, so the fix is just to point the task somewhere the scheduler accepts:</p>
<pre><code class="language-python"># _daemon.py, _render_schtasks_xml — daemon logs via an absolute path, so cwd is irrelevant.
work_dir = Path.home()   # %USERPROFILE%; %APPDATA%\Roaming trips ERROR_DIRECTORY (267)
# ...
#   &lt;WorkingDirectory&gt;{work_dir}&lt;/WorkingDirectory&gt;
</code></pre>
<p><strong>After all three:</strong> a fresh <code>daemon install</code> writes a valid 32-byte key, the task launches (<code>0x00041301</code>), <code>daemon status</code> returns <code>ok: True / WAKE</code>, and <code>capture-hooks install</code> → <code>status: ACTIVE</code> wiring both Claude Code and Desktop. 🎉</p>
<h3>Bonus (test harness, and it touches your #2 file)</h3>
<p><code>tests/test_bench_worktree_resolution.py</code> reads bench scripts with <code>path.read_text()</code> at lines ~110 and ~152 — no <code>encoding="utf-8"</code> — so on Windows it decodes as cp1252 and dies on <code>bench/trajectory.py</code> (<code>UnicodeDecodeError: byte 0x8f</code>). That's the one failure in <code>test_bench_worktree_resolution.py / test_capture_queue.py / test_doctor_lock_probe.py</code> on Windows (1 failed / 45 passed / 19 skipped). One-char fix: <code>read_text(encoding="utf-8")</code>.</p>
<p>One tiny note on #2 itself, not a blocker: the token file uses <code>icacls … /inheritance:d</code>, which <em>copies</em> inherited ACEs rather than stripping them, so it isn't strictly owner-only the way <code>chmod 0600</code> is — <code>/inheritance:r</code> would be closer if that property matters.</p>
<p>Happy to open a small PR with the three daemon fixes (they're independent of the auth work — different files) if that's easier than applying inline. Either way — thanks again for pushing this forward, it's getting really close to a clean Windows experience. 🙏</p>
</body>

</html>

warplayer:
Hey @danielhertz1999-bit — an update from going deeper: I pushed past "the daemon installs" to "iai-mcp is actually *usable* from Claude on Windows" (both the MCP tools and ambient capture), and that surfaced three more issues. All Windows-only, all small fixes. The payoff: explicit recall/save **and** background auto-capture both work on Win11 now.

### A. The node MCP wrapper bridge was never ported (MCP tools unreachable)
The Python IPC got the Windows treatment (TCP + token), but `mcp-wrapper/src/bridge.ts` — the bridge Claude actually talks through — still only dialed the POSIX Unix socket (`~/.iai-mcp/.daemon.sock`), which doesn't exist on Windows. So *every* MCP tool call failed with the hardcoded macOS "daemon not running… launchctl…" error, regardless of daemon state. Mirror `_ipc.py` in the bridge:

```ts
function getDaemonAddress(): string | { host: string; port: number } {
  if (process.platform === "win32") {
    const port = Number.parseInt(
      fs.readFileSync(path.join(os.homedir(), ".iai-mcp", ".daemon.port"), "utf-8").trim(), 10);
    return { host: "127.0.0.1", port };
  }
  return process.env.IAI_DAEMON_SOCKET_PATH ?? path.join(os.homedir(), ".iai-mcp", ".daemon.sock");
}
// on connect (Windows), send the token file's contents as the first line, then RPC.
```
(Same for `emitSessionOpen`, plus a platform-aware error string.) After this, `bridge.start()` connects and an RPC like `schema_list` returns a real response over the TCP+token channel.

### B. `LOCK_SH` isn't shared → ambient capture blocked (your review #3)
With the daemon up, every client store write failed `ConsolidationPendingError`. `_filelock.py`'s `msvcrt.locking` is exclusive-only, so `LOCK_SH` was serviced as `LOCK_EX` — the daemon's WAKE-time "shared" lock blocked all readers. I reimplemented the Windows path on **`LockFileEx`/`UnlockFileEx`** (ctypes): genuine shared locks, EX↔SH conversion via release-then-relock (safe because the `.consolidation-pending` intent file serialises it). Bonus: it locks an explicit byte range, so it also fixes the offset-move and ~10 s-blocking divergences from review #4. `test_hippo_multiprocess`/`_concurrency`/`lock_starvation` all pass on Windows after this.

### C. The kicker — `os.kill(pid, 0)` again, crashing the daemon's tick
Even with B, the daemon never downgraded EX→SH. I instrumented the lifecycle tick: it was dying every iteration at `heartbeat_scanner._is_pid_alive()` → `os.kill(pid, 0)` (`WinError 6`) — the **same POSIX idiom from the `lifecycle_lock` fix in my first review**, in a different file — *before* reaching the downgrade. So the daemon held `EX` forever and B's fix couldn't take effect. Fixed with a psutil probe on Windows, plus the same `os.kill(pid, 0)` in three more files that silently mis-report every pid as dead on Windows (`capture.py`, `doctor/_lifecycle_checks.py`, `cli/_maintenance.py`):

```python
def _is_pid_alive(pid: int) -> bool:
    if pid <= 0:
        return False
    if sys.platform == "win32":
        try:
            import psutil
        except ImportError:
            return True
        return psutil.pid_exists(pid)
    try:
        os.kill(pid, 0)
    except ProcessLookupError:
        return False
    except PermissionError:
        return True
    return True
```

**Verified on Win11:** fresh `daemon install` → daemon downgrades to SHARED on its first WAKE tick → concurrent client captures `inserted` → and in Claude, `memory_capture`/`memory_recall` round-trip. Suite green (lock + heartbeat + capture): 31 passed / 5 skipped.

Happy to send these as a PR (they're independent of the auth-token work). Thanks again — Windows is genuinely usable now. 🙏

danielhertz1999-bit:
## Status update — conflicts resolved + live Windows E2E validation

Rebased onto upstream **v1.1.5** and pushed; the PR is mergeable again. The `write_deferred_captures` conflict was resolved by keeping your pid-collision-safe atomic `.tmp`→`os.replace` rewrite **and** re-applying the `encoding="utf-8"` on both file handles (Windows defaults to cp1252, which corrupts non-ASCII memory content otherwise).

### What landed since the last review (addresses @warplayer's findings)
- **#1 test collection** — the 5 modules that imported `fcntl`/`resource` at top level now use the `_filelock` shim / guard, so collection no longer aborts.
- **#2 `_patch_claude_desktop_config` crash** — now shares the wrapper-missing fallback its Claude Code sibling had; `capture-hooks install` no longer raises after writing hooks.
- **UTF-8 everywhere** — `encoding="utf-8"` on all text I/O across the runtime tree (the cp1252 `UnicodeEncodeError` on `Δ`/emoji/CJK). Verified: cp1252 throws on `Δ`, utf-8 round-trips.
- **`iai_mcp.__main__`** — the package had no `__main__.py`, so the PowerShell hooks' `python -m iai_mcp …` failed with *"No module named iai_mcp.__main__"* and every hook silently no-opped. Added it; also force UTF-8 on stdout/stderr (the recall hook's output path).
- **#4 `_filelock`** — preserve the file offset and emulate POSIX block-until-acquired (msvcrt's `LK_LOCK` gives up after ~10 s).
- **`os.replace` for atomic moves** — fixes the crypto key-rotation `WinError 183`.
- **#3 `LOCK_SH`** — documented as a known divergence (LockFileEx would break the atomic lock *conversion* `hippo/_db.py` relies on).

### Live E2E on Windows 11 (Python 3.12.10, Rust 1.96 MSVC, VS Build Tools)
- `pip install -e .` builds the **Rust native extension + hnswlib** clean; `iai_mcp_native`, `hnswlib`, and `iai_mcp` all import.
- `mcp-wrapper` builds (`tsc` → `dist/index.js`).
- **Core test suite: 627 passed / 68 failed** across 108 modules (~90%). The failures are almost entirely the **POSIX test-harness** assumptions you'd expect — fake daemons on `asyncio.start_unix_server`/`AF_UNIX`, `launchd` re-enable, `signal.SIGKILL`, `0o600`/`geteuid` crypto-mode asserts, `/tmp` paths — not product-logic regressions.
- A cluster of daemon/socket integration tests **hang** under the Windows `asyncio` proactor loop (same "60s socket-bind" stall you hit); these need the test fixtures ported to the TCP-loopback transport.

Happy to split the test-harness port into its own PR if that's easier to review.

_🤖 Validated with [Claude Code](https://claude.com/claude-code) on a real Windows 11 build_

CodeAbra:
Thanks again for this! Looks like #18 is your newer rework of the same Windows port — I'm tracking both. Same note as there: I'm following the work and will merge once the port is complete and validated on a real Windows box. No rush — really appreciate you driving Windows support.

CodeAbra:
Thanks for keeping at this, @danielhertz1999-bit — the rebase onto v1.1.5 and the live Win11 validation (Rust native build, daemon install, the auth-token handshake) are great to see, and the cross-platform shims read cleanly.

I'm ready to merge — the one thing left before I do is the **test-suite port**. The source side is ported, but the tests aren't yet: the POSIX-assuming fixtures (fake daemons on `asyncio.start_unix_server` / `AF_UNIX`, `SIGKILL`, `geteuid`, `/tmp` paths, `0o600` asserts) and especially the daemon/socket integration tests that hang under the Windows proactor loop need porting to the TCP-loopback transport, so the suite actually runs green on Windows. Once that's in and a Win11 run is green, I'll merge.

Splitting the test-harness port into its own commit (as you offered) is totally fine if it's easier to review. Thanks again — this is close.

CodeAbra:
Merged and shipped in **v1.2.0** 🎉 — thank you, @danielhertz1999-bit. Your commit is in the history as the author of the Windows port. I marked Windows as **beta** in the README/badges since the test suite isn't ported yet; once the test-harness port lands and the suite is green on Windows, I'll promote it to fully supported. Huge thanks again — and to @warplayer for the relentless Windows 11 validation that made this mergeable.

(Closed rather than shown as "merged" because it landed via a squash on main; the change itself is live in v1.2.0.)

PR #13, opened 2026-06-18
fix: harden HNSW query + topology schema against small/empty indexes

gardinermichael:
## Summary

Two related issues prevented the memory daemon from accepting **any** inserts on small graphs (N < `top_k`), and made the `topology` MCP tool unusable on graphs below `SIGMA_N_FLOOR`.

### 1. `hippo/_table.py` — `RuntimeError: Cannot return the results in a contiguous 2D array` on every insert

`_ann_to_pandas()` already clamps `k` to `len(_label_map)`, but on boot the label_map and the live hnswlib index can briefly disagree (the rebuild-from-sqlite path in `_initialize_hnsw_index`), and hnswlib raises `Cannot return the results in a contiguous 2D array. Probably ef or M is too small` when asked for more neighbors than the index actually holds.

Because the **pattern-separation gate** (`_store.py:822`) runs `query_similar(k=cfg.top_k)` at every insert, the failure cascaded:

```
capture_turn insert failed
Traceback (most recent call last):
  File ".../iai_mcp/capture.py", line 371, in capture_turn
    store.insert(rec)
  File ".../iai_mcp/store/_store.py", line 306, in insert
    ) = self._pattern_separation_gate_with_hits(record)
  File ".../iai_mcp/store/_store.py", line 822, in _pattern_separation_gate_with_hits
    hits = self.query_similar(list(record.embedding), k=cfg.top_k)
  File ".../iai_mcp/store/_store.py", line 797, in query_similar
    results = q.limit(k).to_pandas()
  File ".../iai_mcp/hippo/_table.py", line 721, in _ann_to_pandas
    labels, distances = db._hnsw.knn_query(self._ann_vector, k=k_clamped)
RuntimeError: Cannot return the results in a contiguous 2D array. Probably ef or M is too small
```

Every queued capture failed → the deferred-captures queue stuck → `N` pinned and the engine effectively stopped ingesting.

**Fix:**
- Also clamp `k` against `db._hnsw.get_current_count()` (the live count, not just the label map).
- Short-circuit on `k_clamped == 0`.
- Defensively wrap `knn_query` in `try/except RuntimeError` → return an empty result and log a warning, so the pattern-separation gate treats it as 'no neighbors found' instead of failing the whole insert.

### 2. `mcp-wrapper/src/tools.ts` — `topology` rejects valid daemon payloads

The `topology` tool's `outputSchema` declared `sigma`, `C`, `L`, and `rich_club_ratio` as `{ type: 'number' }`, but `classify_regime` explicitly returns `sigma=None` when `N < SIGMA_N_FLOOR` (and the other three are similarly undefined on tiny graphs). MCP then rejects the daemon's valid response:

```
MCP error -32602: Structured content does not match the tool's output schema: data/sigma must be number
```

**Fix:** allow `null` alongside `number` for the four fields that are undefined on small graphs.

## Verification

Reproduced the failure locally (fresh-ish store, N=10, default capture hooks active):
- Pre-fix: every deferred-capture drain attempt logged `insert-failed: RuntimeError`; `topology` returned `MCP error -32602`.
- Post-fix: daemon restart drained 8 queued captures (N 10 → 22), no new `RuntimeError` in `launchd-stderr`, and `topology` returns cleanly:
  ```json
  {"C":0.2069,"L":1.8615,"sigma":null,"community_count":1,"rich_club_ratio":0.1364,"N":22,"regime":"insufficient_data"}
  ```

## Test plan

- [ ] Confirm no regression in `tests/test_sigma*.py`
- [ ] Confirm no regression in hippo-table / store tests
- [ ] Verify mcp-wrapper build still passes (`npm run build` in `mcp-wrapper/`)
- [ ] Manual: fresh `~/.iai-mcp/`, install capture hooks, run one session — confirm `iai-mcp topology` and `episodes_recent` both work and N grows


CodeAbra (commented):
Thank you — both diagnoses land. The `_ann_to_pandas` clamp gap during the boot transient is real, and the topology schema mismatch is exactly why `classify_regime` returns `None` for sigma on small graphs.

The fix itself looks good. One small follow-up if you have the bandwidth: it'd be great to land a tests-only commit covering the three new edge cases:

1. `_ann_to_pandas` with `k_clamped == 0` returning empty without calling `knn_query`.
2. A simulated `RuntimeError` from `knn_query` → empty result, no re-raise.
3. Topology RPC parsing with `sigma == null` (and the other three fields null).

If those land in the next day or so, great — otherwise we'll merge as-is and pick the tests up on our side, since the cascade prevention is the priority and shouldn't wait. Thanks again!

CodeAbra:
Thanks @gardinermichael — both diagnoses were spot on. The HNSW small/empty-index hardening you proposed (clamp `k` to the live index count and short-circuit `k_clamped == 0` before `knn_query`) has since landed on main via the consolidation/recall-correctness work, and the topology schema now tolerates null `sigma` on small graphs. So this is already covered in v1.1.5/v1.1.6 — closing as resolved. Really appreciate the careful report and fix; please keep them coming.

CodeAbra:
Small correction to my note above: the HNSW `k_clamped` short-circuit is on main, but the topology-schema part (accepting a null `sigma` on small graphs) is not yet shipped — it's a one-line wrapper-schema change we're tracking as a tiny follow-up. Thanks again @gardinermichael; the substantive fix is in.

PR #14, opened 2026-06-19
Fix episodic capture timestamp matcher and duplicate-capture race

joerybka:
## Summary
  - Fixes `_find_transcript_ts`'s content-hash fallback, which compared a record's `literal_surface` against `obj.get("text")`/`obj.get("content")` at the top level of each transcript JSON line — but real Claude Code transcript entries nest the turn text under `message.content`. Any
  record without a stored `source_uuid` could never be matched, leaving its `created_at` stuck at the collapsed value no matter how many times the repair migration ran.
  - Fixes a race in `capture_turn()`'s dedup: the daemon's RPC dispatch runs each request on its own thread (`asyncio.to_thread`), and Stop-hook captures replay the full transcript on every firing with no incremental offset tracking. A session with many sub-agent/fork completions can
  fire its Stop hook a dozen+ times in under an hour, queueing near-identical deferred-capture files; concurrent drains of those files could both see an idem tag as absent and both insert, producing true duplicate episodic records. Confirmed live: one session fired Stop 16 times in 45
  minutes and produced 2-3x duplicates for several turns.
  - Adds `iai-mcp migrate --dedupe-episodic`, a one-off cleanup migration that groups existing records by idem tag and tombstones every duplicate but the first (soft-delete only — `literal_surface`/`provenance`/embeddings untouched).
  - Fixes `doctor`'s collapsed-timestamp check, which was missing the `tombstoned_at IS NULL` filter the migration's own candidate query already had.
  - Adds `bench.capture_dedup_lock`, a new bench harness for the lock fix's overhead.

  Four atomic commits, one logical change each.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / tooling

## Affected areas

- [x] Capture path
- [ ] Recall / retrieval
- [ ] Daemon lifecycle / FSM
- [x] Bench harness
- [x] CLI / doctor
- [ ] Other: ___

## Testing

- [x] `pytest` passes locally
- [x] `ruff check src/ tests/` clean

### Runs
- [x] `pytest tests/test_migrate.py tests/test_capture.py tests/test_doctor_hippo_rows.py` — 44 pass, including new regression tests for the matcher fix, the concurrency race (verified it fails without the lock: `max_in_flight=6`), and the dedupe migration
  (insert/idempotent/dry-run/event).
- [x] Full repo suite: 3458 passed / 125 skipped / 1 xfailed / 3 failed. All 3 failures pre-existing and unrelated: 2 are the documented test-pollution-sensitive flakes (pass on rerun, per CONTRIBUTING.md), 1 is an HNSW recall-parity test that reproduces identically on a clean
  `upstream/main` checkout.
- [x] `ruff check` on every changed/added file: zero new violations (diffed against a clean `upstream/main` baseline — all pre-existing hits are unchanged; the new bench file's `E402`s match the same sys.path-then-import pattern already present in every other `bench/*.py`).
- [x] Applied against a live store: 1007 duplicate records across 168 groups tombstoned; `iai doctor`'s collapsed-timestamp check now passes clean.

## Benchmarks

  This touches `capture` (the `_CAPTURE_DEDUP_LOCK` addition), so per the template:

  - Bench command run: `python -m bench.capture_dedup_lock`
  - Before (no lock, 3 rounds, median): solo capture_turn p50=158.95ms / mean=162.12ms / p95=164.31ms; concurrent throughput (8 threads × 20 distinct captures) = 16.7 captures/sec
  - After (with lock, 3 rounds, median): solo capture_turn p50=158.27ms / mean=160.11ms / p95=163.62ms; concurrent throughput = 17.1 captures/sec

  Solo latency is unchanged (zero lock contention by construction). Concurrent throughput shows no regression — the ~2% delta is within run-to-run noise, since the embedding call (the actual bottleneck, ~158ms) sits outside the lock; only the cheap dedup-check + insert is now
  serialized, which was already largely serialized by the existing `_conn_lock` and the synchronous wait on the async write queue.

  This PR doesn't touch retrieval, ranking, or embedding code, so the retrieval-side benches aren't expected to move. Ran them anyway as a sanity check:
  - `bench.tokens --wake-depth minimal` (real store, read-only, post-`--dedupe-episodic`): `fresh=392 warm=[16,16,16] steady_ok=true fresh_ok=true`
  - `bench.verbatim` (isolated store, n=20): `accuracy=1.0 passed=true`
  - `bench.neural_map --n 100 --iterations 5` (isolated store, reduced scale): `p50=61.4ms p95=66.3ms passed=true`, well under the 100ms gate

  Skipped `bench.longmemeval_blind` (the 500-question full run) — it has no code-path overlap with this PR and is disproportionately heavy for a write-path-only change.

### DIFF
--- a/bench/capture_dedup_lock.py
+++ b/bench/capture_dedup_lock.py
@@ -1,0 +1,134 @@
+"""Capture dedup-lock overhead: capture_turn() solo latency and concurrent throughput.
+
+Quantifies the cost of _CAPTURE_DEDUP_LOCK (the fix for the daemon's
+concurrent-thread capture_turn() dedup race -- see capture.py). All
+captures used here have distinct content, so every call falls through to
+an actual insert; the question this bench answers is how much serializing
+the dedup-check-through-insert critical section costs, both with zero
+contention (solo) and under real thread contention (concurrent).
+"""
+from __future__ import annotations
+
+import argparse
+import statistics
+import sys
+import tempfile
+import threading
+import time
+from pathlib import Path
+from uuid import uuid4
+
+_SRC_PATH = str(Path(__file__).resolve().parent.parent / "src")
+if _SRC_PATH not in sys.path:
+    sys.path.insert(0, _SRC_PATH)
+
+import os
+if not os.environ.get("IAI_MCP_CRYPTO_PASSPHRASE"):
+    os.environ["IAI_MCP_CRYPTO_PASSPHRASE"] = (
+        "iai-mcp-bench-falsifiability-deterministic-2026"
+    )
+
+from iai_mcp.capture import capture_turn
+from iai_mcp.store import MemoryStore
+
+SESSION_ID = "bench-capture-dedup-lock"
+MIN_TEXT = (
+    "with enough length to clear the MIN_CAPTURE_LEN floor for a realistic capture"
+)
+
+
+def _solo_latency(store: MemoryStore, n: int) -> dict:
+    samples_ms: list[float] = []
+    for i in range(n):
+        t0 = time.perf_counter()
+        capture_turn(
+            store,
+            cue=f"bench solo turn {i}",
+            text=f"Capture dedup-lock bench solo turn {i} {MIN_TEXT}.",
+            tier="episodic",
+            session_id=SESSION_ID,
+            role="user",
+            source_uuid=str(uuid4()),
+        )
+        samples_ms.append((time.perf_counter() - t0) * 1000.0)
+    samples_ms.sort()
+    return {
+        "n": n,
+        "p50_ms": statistics.median(samples_ms),
+        "mean_ms": statistics.mean(samples_ms),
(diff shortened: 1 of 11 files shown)


CodeAbra (approved):
Excellent, careful PR — thank you. The four commits split cleanly and each one stands on its own:

- The timestamp matcher fix in commit 1 handles transcript bodies nested under `message.content`, which the previous top-level fallback missed.
- The dedup lock in commit 2 is the right shape — scoped to the episodic find-then-insert window, doesn't block the rest of the capture path.
- The doctor `tombstoned_at IS NULL` filter in commit 3 is a clean one-liner.
- The `--dedupe-episodic` migration in commit 4 is conservative (soft-delete only, smallest-id survivor, `literal_surface` never touched) and idempotent.

Tests look solid. Merging — and keeping your four-commit split intact. Thank you for the contribution!

PR #19, opened 2026-06-20
fix(capture): embed message content, not the positional cue label

Marsu6996:
## Problem

`capture_transcript()` calls `capture_turn(cue=f"session {session_id} turn {seen}", text=<message>)`, and `capture_turn()` embedded `cue or text`. Because the cue is always a truthy positional label, **every episodic capture's vector is the embedding of `"session <uuid> turn <n>"`, not the message content**. `literal_surface` still stores the real text (no data loss), but the vector index does not represent the content.

### Impact (measured on a real ~2400-record store)
- Stored vectors collapse: cosine between *random, unrelated* captures ≈ **0.91** (vs ≈ **0.69** when the same texts are embedded fresh), because every vector is the embedding of a near-identical label. Across 200 sampled records: `cos(stored_vector, embed(cue)) = 1.000` vs `cos(stored_vector, embed(literal_surface)) ≈ 0.60`.
- The pattern-separation gate only links neighbours within the band `[link_threshold, near_dup_threshold)` = `[0.70, 0.92)`; with almost everything ≥ 0.92 (treated as near-dup), virtually no edges form → near-edgeless similarity graph → 1 trivial community → degraded `rich_club`/topology → `crisis_mode`.
- Semantic recall returns flat / off-topic hits regardless of the query (it cannot rely on the collapsed vector space).

This affects **all users** (the cue is a positional id, not semantic), and is especially visible for non-English content.

## Fix

Embed `text` (the content) in `capture_turn()`. The cue stays a provenance label only. Fixing this single chokepoint covers the transcript, deferred-drain and recovery paths at once.

Re-embedding existing records from their intact `literal_surface` (e.g. via the reembed migration) restores the vector space: random-pair cosine drops 0.91 → 0.69 and recall returns relevant results again.

## Test

`test_capture_turn_embeds_content_not_cue` monkeypatches the embedder to record its input and asserts `capture_turn()` embeds the message text, never the cue. It fails on the pre-fix code, locking the regression.

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -277,7 +277,11 @@ def capture_turn(
     from iai_mcp.events import TELEMETRY_EMBED_NATIVE_FAILURE, write_event
 
     try:
-        emb = embedder_for_store(store).embed(cue or text)
+        # Embed the CONTENT, never the cue. capture_transcript / deferred-drain pass
+        # a positional cue ("session <id> turn <n>"); embedding that collapsed every
+        # stored vector onto near-identical label embeddings, destroying graph
+        # structure and semantic recall. The cue stays a provenance label only.
+        emb = embedder_for_store(store).embed(text)
     except Exception as exc:
         write_event(
             store,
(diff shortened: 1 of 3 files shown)


CodeAbra:
Shipped in v1.1.3 — your fix is in, credited as co-author on the commit. Embedding the cue label instead of the message content was a real recall-killer; thanks for the careful diagnosis and the before/after measurement. We also added `iai-mcp migrate --reembed-from-text` so existing stores can repair vectors written before the fix. Closing with thanks!

PR #21, opened 2026-06-21
Port from macOS to Linux (Fedora 44)

MoppelMat:
- Bump Python >=3.11,<3.15 (Fedora 44 / Python 3.14)

    Fix systemd unit template (remove StartLimit keys)
    Add install-linux.sh one-command setup script
    Fix test_daemon.py and test_bench_worktree_resolution.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 <p align="center">
   <img src="https://img.shields.io/badge/release-v1.1.2-1f6feb?style=flat-square" alt="Release v1.1.2">
   <a href="LICENSE"><img src="https://img.shields.io/badge/license-MIT-1f6feb?style=flat-square" alt="License: MIT"></a>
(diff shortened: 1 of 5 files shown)


CodeAbra:
Thanks @MoppelMat — landed the Linux install helper and the systemd `[Service]` StartLimit cleanup in v1.1.7 (the StartLimit keys are only honored in `[Unit]`, so dropping the dead ones removes the modern-systemd warning).

I held back the `requires-python` bump to `<3.15` (and the 3.13/3.14 badge) for now: I built the package on 3.14 and it fails — `pyarrow` has no 3.14 wheel yet, so `pip install` falls back to a from-source Arrow build and breaks. Once pyarrow ships 3.14 wheels (3.13 likely already fine), happy to widen it in a follow-up — a focused PR that only touches `requires-python` + the badge would be easy to take. Really appreciate the contribution.

PR #22, opened 2026-06-21
fix(memory): consolidation & recall correctness — live crisis graph, score clamp, reflection embedding, crash recovery

Marsu6996:
## Summary

This is the consolidation- and recall-correctness half of #17 (the daemon
WAKE/idle CPU-storm half is the companion PR). After a healthy store accumulates
soft-deleted / deduped records, the sleep pipeline rebuilds its topology over the
**tombstoned** nodes too, so `rich_club` sits just under its crisis floor and
re-arms `crisis_mode` every cycle; meanwhile a degraded daemon serves recall as
flat `1.000` scores. Each commit fixes one root cause behind those symptoms, on
top of `v1.1.4`.

### What changed (one logical change per commit)
- **`fix(capture)` — stop re-embedding seen turns, close the dedup race.** Active
  sessions re-drain the whole transcript every turn, so the eager embed at the top
  of `capture_turn` re-ran the GIL-bound Rust matmul for every already-stored turn
  just to discard it on the idem-tag check. Embedding is now deferred behind a
  memoized `_compute_embedding()` that runs at most once and only for genuinely new
  turns; `flush_record_buffer` before the idem lookup under `_CAPTURE_DEDUP_LOCK`
  closes a check-then-insert race that produced live duplicate records.
- **`fix(recall)` — clamp the displayed score to [0,1], rank on unclamped
  `sort_score`.** Multiplicative boosts (trigram×2, FTS×3, valence) push the
  internal score past 1.0, so a served hit could surface a "confidence" > 1 and a
  degraded state surfaced flat `1.000`. The *displayed* score is clamped at
  serialization while ranking uses a new, unclamped `MemoryHit.sort_score`; the
  stale-downweight keeps the two in lock-step. **Ordering is provably unchanged**
  (regression test included).
- **`fix(consolidation)` — build crisis topology on the live graph; demote
  `rich_club`.** The essential-variable tracker and crisis recluster rebuilt over
  ALL records including tombstoned ones, so communities / centrality / `rich_club`
  were computed on thousands of dead nodes (`rich_club ~0.019`, just under the 0.02
  floor → re-arms crisis on a healthy store). A shared `build_live_graph()`
  (tombstone-filtered nodes, live-only edges) feeds both paths, and
  `rich_club_coefficient` is demoted from a crisis *trigger* to a diagnostic-only
  signal (`edge_density` and `community_count` remain triggers). The runtime-graph
  tombstone guard is hardened to `pd.isna()` so a `NaT`/`NA` `tombstoned_at` on a
  reembedded `datetime64` column reads as LIVE instead of collapsing the graph.
- **`fix(dmn)` — embed the reflection surface text instead of a zero placeholder.**
  Daily reflections wrote an all-zero embedding that was never re-embedded, leaving
  them permanently unretrievable and feeding zero vectors into the scoring matmul.
  Now embeds `literal_surface`; on native failure falls back to a zero vector
  flagged `embedding_pending=1` for deferred reembed.
- **`fix(sigma)` — bound small-worldness with a node-count ceiling.** Adds
  `SIGMA_N_CEIL` (default 20000, env `IAI_MCP_SIGMA_N_FLOOR` / `..._CEIL`) returning
  `None` above it so the unbounded random-reference compute can't spin a core.
- **`fix(watchdog)` — monitor a retried-but-wedged sleep cycle for staleness.** The
  staleness check only monitored `attempt == 1`, so a retried-and-still-wedged
  cycle (`attempt >= 2`) was short-circuited and ignored. Gates on `attempt < 1`
  instead (excluding `bool`), so every genuine running attempt is monitored.
- **`feat(daemon)` — surface a repeated store-empty count failure as telemetry.**
  The companion PR stops a transient count failure from parking the tick but left
  the condition invisible; this emits a best-effort `store_empty_check_failed`
  warning event (buffered, never raises).
- **`fix(hippo)` — decrypt `literal_surface` before reembedding pending rows.** On
  an encrypted store, `reembed_pending_rows` embedded the raw `iai:enc:v1:`
  ciphertext, producing a garbage vector for every `embedding_pending=1` row. Now
  decrypts via the existing `_decrypt_record_field` (a no-op on a plaintext store);
  a decrypt failure leaves the row pending for retry rather than poisoning it.
  *Distinct from the v1.1.4 reembed fix (`04e62e2`):* that repaired the **migration**
  path (`migrate/_reembed_from_text.py::migrate_reembed_from_text`); this fixes the
  **runtime/daemon** path (`hippo/_db.py::reembed_pending_rows`), which v1.1.4 does
  not touch — so the two are complementary, not redundant.
- **`fix(daemon)` — recover cleanly from a crash mid-SLEEP at boot.** A daemon
  killed mid-SLEEP leaves `lifecycle_state.json` at `SLEEP` with
  `sleep_cycle_progress=None` (incoherent); resuming wedges the daemon so it never
  reaches the recluster that clears crisis. A pure `_normalize_boot_lifecycle_state`
  resets exactly that case to a clean WAKE and clears the stale crisis flags at boot.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / tooling

## Affected areas

- [x] Capture path
- [x] Recall / retrieval
- [x] Consolidation / sleep cycles
- [x] Daemon lifecycle / FSM
- [x] Storage / encryption at rest
- [ ] Bench harness
- [ ] CLI / doctor
- [ ] Other: ___

## Testing

- [x] `pytest` passes locally — full default gate
  (`pytest -m "not perf and not slow and not live"`, 3538 tests), rebased on
  `v1.1.4`: **3514 passed, 33 skipped, 1 xfailed, 1 failed**. The single failure
  (`test_rendered_plist_contains_fd_floor`) is **pre-existing — it fails
  identically on a clean `v1.1.4` checkout** and is unrelated to this branch.
  (Some `test_doctor_*` rows are environment-flaky on the dev machine — a large
  system `subprocess` output decoded as strict UTF-8 — and pass on a clean run.)
- [x] `ruff check src/ tests/` — no new findings vs the `v1.1.3` baseline on the
  touched files (the repo ships no ruff config).
- [x] New tests added for changed behaviour — score-clamp order preservation,
  tombstone/live-graph filtering on `NaT`/`NA` columns, `dmn` embedding norm,
  sigma ceiling + env override, watchdog stale gate, encrypted-store reembed
  decrypt, and crash-mid-SLEEP boot normalization. (The `capture` drain/dedup
  commit currently relies on the existing capture tests + a live-store check — a
  dedicated unit test would strengthen it; happy to add if you'd like one here.)

## Benchmarks

The recall change is **display-only and order-preserving**: ranking moves to an
unclamped `sort_score`, the clamp touches only the serialized number, and the
regression test asserts identical ordering — so LongMemEval-S is unaffected by
construction. The `dmn` fix makes daily reflections retrievable again (previously
zero-vector), which can only help recall on reflection cues. A blind LongMemEval-S
run (`python -m bench.longmemeval_blind --split S`) can be attached on request; it
is not run by default to keep the blind split blind.

- Bench command run: _pending — blind LongMemEval-S queued before merge if desired_
- Before:
- After:

## Notes for reviewers

- **Relates to #17** (intentionally not "Fixes": the companion WAKE/idle PR carries
  the closing keyword, since you flagged the CPU-storm half as the gate for closing
  #17). This PR addresses the consolidation/recall symptoms — crisis loop from a
  tombstoned-polluted graph, flat `1.000` served recall — at their source.
- **Stacked on the WAKE/idle PR.** Several files build on it (`capture.py` uses
  `is_drain_in_progress`, the runtime-graph tombstone guard, the store-empty path),
  so this is best reviewed/merged after it. Rebasing onto `main` once that lands is
  a clean fast-forward.
- **Complements the crisis auto-expiry** (`crisis_mode_since_ts`, shipped v1.1.3)
  rather than duplicating it: the 72 h timeout clears a *coherent* crisis; this
  removes the false trigger so crisis is not (re-)armed on a healthy store in the
  first place.
- **Rebased on `v1.1.4`.** Our files are disjoint from v1.1.4's changes (reembed
  migration + analytics), so the rebase was conflict-free; the full gate above was
  run on the v1.1.4-based branch.
- `rich_club_coefficient` is no longer a crisis trigger (kept as a diagnostic
  event field `is_crisis_trigger`). New env knobs: `IAI_MCP_SIGMA_N_FLOOR` /
  `IAI_MCP_SIGMA_N_CEIL`. New `MemoryHit.sort_score` field defaults to `None`
  (backward compatible — callers fall back to `score`).

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -1,6 +1,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import hashlib
 import json
 import logging
@@ -40,6 +41,46 @@ MAX_CAPTURE_LEN = 8000
 # This serializes that sequence so a tag can never be checked-as-absent by
 # two threads before either has inserted it.
 _CAPTURE_DEDUP_LOCK = threading.Lock()
+
+# Drain-in-progress accounting for the daemon's lifecycle idle countdown.
+#
+# The wrapper heartbeat file is only ONE signal that the daemon is busy. A
+# deferred-capture drain (triggered by real RPC traffic, kicked off at the
+# drowsy edge, or run on startup) can keep writing to the store long after the
+# heartbeat file has gone stale. If the idle countdown only watched the
+# heartbeat, it would force the FSM toward SLEEP -- which escalates to an
+# EXCLUSIVE store lock -- while these drain threads are still hammering the
+# store, causing lock contention. The daemon consults `is_drain_in_progress`
+# so an in-flight drain holds the idle countdown open. A depth counter (rather
+# than a boolean) keeps this correct under concurrent/overlapping drains across
+# the several threads that asyncio.to_thread dispatch spins up.
+_DRAIN_IN_PROGRESS_LOCK = threading.Lock()
+_drain_in_progress_depth = 0
+
+
+@contextlib.contextmanager
+def _drain_in_progress_guard():
+    """Mark a deferred-capture drain as in flight for its whole duration."""
+    global _drain_in_progress_depth
+    with _DRAIN_IN_PROGRESS_LOCK:
+        _drain_in_progress_depth += 1
+    try:
+        yield
+    finally:
+        with _DRAIN_IN_PROGRESS_LOCK:
+            _drain_in_progress_depth -= 1
+
+
+def is_drain_in_progress() -> bool:
+    """True while at least one deferred-capture drain is running.
+
+    The daemon's lifecycle idle countdown reads this so it does NOT advance the
+    FSM toward SLEEP (and the EXCLUSIVE store lock it takes) while drain threads
+    are still writing to the store.
+    """
+    with _DRAIN_IN_PROGRESS_LOCK:
+        return _drain_in_progress_depth > 0
+
 
 FAILED_MAX_ATTEMPTS: int = 3
 FAILED_BACKOFF_BASE_SEC: float = 60.0
@@ -711,6 +752,17 @@ def write_deferred_captures(
 
 
 def drain_deferred_captures(store: MemoryStore) -> dict[str, int]:
+    """Drain crashed/orphaned deferred-capture files into the store.
+
+    Marks itself in-progress for the daemon idle countdown so the FSM does not
+    advance toward SLEEP (EXCLUSIVE store lock) mid-drain. See
+    `is_drain_in_progress`.
+    """
+    with _drain_in_progress_guard():
+        return _drain_deferred_captures_impl(store)
+
+
+def _drain_deferred_captures_impl(store: MemoryStore) -> dict[str, int]:
     deferred_dir = Path.home() / ".iai-mcp" / ".deferred-captures"
     log_dir = Path.home() / ".iai-mcp" / "logs"
     log_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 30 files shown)


CodeAbra (approved):
Thanks — this carries the consolidation/recall-correctness root causes for the crisis loop: tombstoned-record exclusion from the runtime graph, crisis topology on the live graph, score clamp, reflection-embedding + crash recovery. Reviewed the diff, security-clean, CI green. Merging with credit to you.

Issue #26, opened 2026-06-27
capture-hooks install rewrites Claude MCP config to packaged wrapper which cannot resolve @modelcontextprotocol/sdk


Marsu6996:
## Summary

`iai-mcp capture-hooks install` rewrites `~/.claude.json` and `~/Library/Application Support/Claude/claude_desktop_config.json` so that the `iai-mcp` MCP server is launched via the wrapper shipped inside the wheel:

```
/Users/<user>/<env>/.venv/lib/python3.11/site-packages/iai_mcp/_wrapper/index.js
```

That directory ships the wrapper JS files but no `package.json` and no `node_modules/`. As soon as Claude Desktop or Claude Code tries to spawn it, Node fails with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@modelcontextprotocol/sdk' imported from
  .../site-packages/iai_mcp/_wrapper/index.js
```

and the MCP entry shows up as `failed` in the Claude Desktop developer panel.

## Repro

1. Clone the repo, set up the venv, `pip install .` (so the wheel is installed into site-packages with the bundled `_wrapper/` JS but no node deps).
2. Run `iai-mcp capture-hooks install`.
3. Restart Claude Desktop.
4. Open Settings → Developer → MCP. The `iai-mcp` entry is `failed`; the log shows `ERR_MODULE_NOT_FOUND: @modelcontextprotocol/sdk`.

## Why it works on dev installs that build the wrapper separately

In a dev layout, the actually-runnable wrapper lives at `mcp-wrapper/dist/index.js` next to its `node_modules/` (after `npm install && npm run build` in `mcp-wrapper/`). The pre-v1.2.0 Claude configs on at least this machine pointed there and worked. The new `capture-hooks install` replaces that path with the in-wheel one and breaks startup.

## Affected versions

Observed on **v1.2.0** (latest at the time of writing). Did not check whether the same rewrite happened in v1.1.x — happy to confirm if helpful.

## Expected behaviour

Either:

* ship the wrapper's npm dependencies inside the wheel (so the in-wheel `_wrapper/index.js` is genuinely runnable as-is), **or**
* have `capture-hooks install` detect a usable `mcp-wrapper/dist/index.js` next to the venv root and prefer that path, **or**
* document that `capture-hooks install` is intended only for self-contained installs and skip the MCP config rewrite when the in-wheel wrapper has no `node_modules/`.

## Local workaround

[...]

PR #28, opened 2026-06-27
Fix capture-hooks install wiring a non-runnable in-wheel MCP wrapper (#26)

danielhertz1999-bit:
Fixes the first half of #26.

## The bug

`_resolve_wrapper_path()` returns the wheel-bundled wrapper (`iai_mcp/_wrapper/index.js`) as soon as it **exists**, ahead of a working `mcp-wrapper/dist/index.js`. But the in-wheel copy ships the JS with **no `package.json` and no `node_modules/`**, so when Claude spawns it Node throws:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@modelcontextprotocol/sdk'
  imported from .../site-packages/iai_mcp/_wrapper/index.js
```

and the `iai-mcp` MCP entry shows as **failed**. `capture-hooks install` overwrites a previously-working dev path with this broken one.

## The fix

Prefer a wrapper whose npm deps **actually resolve**. Node resolves bare imports like `@modelcontextprotocol/sdk` by walking parent dirs for a `node_modules` containing the package, so the resolver now does the same: it collects the candidate paths in preference order and returns the first one with a reachable `node_modules/@modelcontextprotocol/sdk`. The depless in-wheel wrapper is skipped in favour of `mcp-wrapper/dist` sitting next to its `node_modules`.

When a wrapper exists but its deps aren't installed (e.g. `pip install .` with no `npm` build and no dev tree), it now raises `FileNotFoundError` with a clear *"build it / set `IAI_MCP_WRAPPER_PATH`"* message instead of silently wiring a broken entry — callers already route that through the existing placeholder/warning path (`_iai_entry_or_placeholder`), so install degrades gracefully rather than producing a `failed` MCP server.

## Verification

Unit-tested the resolver against both layouts: in-wheel-without-`node_modules` is correctly treated as not-runnable and skipped; `dist`-with-`node_modules` is preferred. No behavioral change when the in-wheel wrapper is genuinely runnable (i.e. once the wheel bundles its deps).

## Not included here

The second half of #26 (SessionStart hook re-registration / lack of a `--components` opt-out) is a separate ergonomics change — happy to send a follow-up adding a `--components=stop,turn` style flag if you'd like that approach.

### DIFF
--- a/src/iai_mcp/cli/_capture.py
+++ b/src/iai_mcp/cli/_capture.py
@@ -337,6 +337,24 @@ def _turn_hook_paths() -> tuple:
     return src, dst
 
 
+def _wrapper_deps_resolvable(index_js: Path) -> bool:
+    """Whether ``index_js``'s npm dependencies actually resolve from its location.
+
+    Node resolves bare imports like ``@modelcontextprotocol/sdk`` by walking
+    parent directories for a ``node_modules`` that contains the package. The
+    wrapper bundled inside the wheel (``iai_mcp/_wrapper/index.js``) ships the JS
+    but no ``node_modules``, so spawning it fails with
+    ``ERR_MODULE_NOT_FOUND: @modelcontextprotocol/sdk``. This lets resolution
+    prefer a wrapper that is genuinely runnable (e.g. ``mcp-wrapper/dist``
+    sitting next to its ``node_modules``) over one that merely exists.
+    """
+    sentinel = Path("node_modules") / "@modelcontextprotocol" / "sdk"
+    for parent in index_js.parents:
+        if (parent / sentinel).exists():
+            return True
+    return False
+
+
 def _resolve_wrapper_path() -> Path:
     import iai_mcp as _pkg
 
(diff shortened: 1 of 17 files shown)


Marsu6996 (approved):
Tested locally against my v1.2.0 install (the one reported in #26). Confirmed:

- On my real environment, the **old** `_resolve_wrapper_path()` returned the in-wheel `_wrapper/index.js` (which has no `node_modules` alongside it) — that's exactly the path that fed `~/.claude.json` with a non-runnable wrapper and made the `iai-mcp` MCP entry fail to spawn with `ERR_MODULE_NOT_FOUND: @modelcontextprotocol/sdk`.

- With this patch applied, `_wrapper_deps_resolvable()` correctly mirrors Node's bare-import resolution (walks parents looking for `node_modules/@modelcontextprotocol/sdk`). Verified on three layouts: isolated `index.js` (returns False), sibling `node_modules` (True), and `dist/index.js` with `../node_modules` (True — the realistic `mcp-wrapper` editable case).

- On my install where no candidate has resolvable deps, the patched code now raises `FileNotFoundError` with the actionable "build it / set `IAI_MCP_WRAPPER_PATH`" message instead of silently writing a broken MCP entry. Callers already route that through the placeholder/warning path, so end-to-end the install fails loud rather than fails silent.

Resolves the first half of #26 cleanly. Approving.

CodeAbra:
Merged into main (squash) — thanks @danielhertz1999-bit! This fixes the install half of #26 (the in-wheel wrapper that couldn't resolve the SDK). Landed via squash, hence closed not merged.

PR #29, opened 2026-06-27
capture-hooks install: add --components opt-out (SessionStart re-registration) (#26)

danielhertz1999-bit:
Addresses the second half of #26.

## The problem

`capture-hooks install` always (re)registers all three capture hooks — `Stop`, `UserPromptSubmit`, and the `SessionStart` recall/prefix hook — and there's no way to opt out. A user who had deliberately removed the `SessionStart` prefix injection (too noisy at every startup/resume/clear/compact) but kept `Stop` + `UserPromptSubmit` capture got the `SessionStart` entry **silently put back** on the next `install`, with no `--force` to gate it and no flag to skip it.

## The change

Add a `--components` flag to `capture-hooks install`:

- **Default `stop,turn,recall`** — identical to today's behavior, so existing installs are unaffected.
- **`--components=stop,turn`** — installs ambient capture **without** wiring the `SessionStart` recall hook. An excluded component's template copy *and* its `settings.json` registration are both skipped.
- Unknown component names are rejected with a clear message and exit 2.

```
capture components to install (default: all). Choices:
  stop   — end-of-session capture (Stop hook)
  turn   — per-prompt capture (UserPromptSubmit hook)
  recall — SessionStart prefix injection
```

`status` and `uninstall` already report/remove each hook independently, so no change was needed there. Refactored the repeated template-copy / command-string construction into small local helpers so the three component blocks read in parallel.

## Verification

Ran the installer against a redirected temp `$HOME`:
- `--components=stop,turn` → `settings.json` has `Stop` + `UserPromptSubmit`, **no `SessionStart`**.
- default → all three present (unchanged).
- `--components=stop,bogus` → exits 2 with `unknown capture component(s): bogus`.

## Note

This doesn't try to auto-detect "deliberately removed" (which isn't reliably distinguishable from "never installed" without a stamped marker) — it gives the explicit opt-out you suggested instead. Pairs with #28 (the in-wheel-wrapper resolution fix) to close out #26.

### DIFF
--- a/src/iai_mcp/cli/__init__.py
+++ b/src/iai_mcp/cli/__init__.py
@@ -607,9 +607,23 @@ def _build_parser() -> argparse.ArgumentParser:
         help="install/uninstall/status the Claude Code Stop hook for ambient session capture",
     )
     ch_sub = ch.add_subparsers(dest="capture_hooks_cmd", required=True)
-    ch_sub.add_parser("install",
-                      help="copy Stop hook to ~/.claude/hooks/ and register in settings.json"
-                      ).set_defaults(func=cmd_capture_hooks_install)
+    ch_install = ch_sub.add_parser(
+        "install",
+        help="copy capture hooks to ~/.claude/hooks/ and register in settings.json",
+    )
+    ch_install.add_argument(
+        "--components",
+        default="stop,turn,recall",
+        metavar="LIST",
+        help=(
+            "comma-separated capture components to install (default: all). "
+            "Choices: stop (end-of-session capture), turn (per-prompt capture), "
+            "recall (SessionStart prefix injection). "
+            "e.g. --components=stop,turn installs capture WITHOUT the SessionStart "
+            "recall hook."
+        ),
+    )
+    ch_install.set_defaults(func=cmd_capture_hooks_install)
     ch_sub.add_parser("uninstall",
                       help="remove the Stop hook and its settings.json entry"
                       ).set_defaults(func=cmd_capture_hooks_uninstall)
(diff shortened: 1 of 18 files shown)


Marsu6996 (approved):
Tested locally — covers exactly the scenario I hit on v1.2.0 (the second half of #26).

**Parser unit tests** (`_parse_hook_components`): default / empty / whitespace all return the full set; `stop,turn` returns `{stop,turn}`; case-insensitive matching works; unknown components (`foo`, `stop,foo`, `stop,turn,nope`) all raise `ValueError` with a clear "valid: stop, turn, recall" message → exit code 2 via `cmd_capture_hooks_install`. ✓

**Integration test** in a sandboxed `$HOME`:

1. `--components=stop,turn,recall` (default) — installs all three hook templates and registers Stop, UserPromptSubmit, and SessionStart in `settings.json`. Unchanged behavior preserved. ✓
2. `--components=stop,turn` — installs only `iai-mcp-session-capture.sh` and `iai-mcp-turn-capture.sh`; `iai-mcp-session-recall.sh` is **not** copied; `settings.json` ends with `Stop` and `UserPromptSubmit` only, **no `SessionStart` entry**. Explicit `skipped: SessionStart recall hook (not in --components)` log line. ✓
3. **Regression scenario from #26** — initial install with all components, then manual removal of `SessionStart` from `settings.json`, then `install --components=stop,turn` re-run: `SessionStart` stays out. No silent re-injection. ✓
4. `--components=stop,foo,recall` — fails fast with `ERROR: unknown capture component(s): foo. valid: stop, turn, recall` and exit code 2. ✓

The refactor into `_install_template` / `_hook_command` helpers keeps the three component blocks symmetric and readable. Approving.

CodeAbra:
Merged into main (squash) — thanks @danielhertz1999-bit! The --components opt-out addresses the second half of #26.

PR #30, opened 2026-06-27
tests: fix NameError in daemon_sock / fake_daemon fixtures (macOS CI)

danielhertz1999-bit:
The **build & test (macOS)** gate is currently red on `main` (incl. the [Release v1.2.0 run](https://github.com/CodeAbra/iai-personal-memory-engine/actions/runs/28267001306)). This fixes the most clear-cut part of it.

## What's broken

`daemon_sock` (test_mcp_tools.py) and `fake_daemon` (test_socket_disconnect_reconnect.py) both reference `tmp_path` in the fixture body without it being a parameter:

```python
@pytest.fixture(scope="module")
def daemon_sock() -> "Path":
    sock_dir = tmp_path / "sock"   # NameError: name 'tmp_path' is not defined
```

So every test using them errors at setup — 5 errors on the macOS gate. (`daemon_sock` is also `scope="module"`, where the function-scoped `tmp_path` wouldn't be injectable anyway.)

## Fix

Bind the control socket under a short `tempfile.mkdtemp(prefix="iai-sock-")` dir (the same pattern as the existing `short_socket` conftest fixture). This:
- removes the undefined-name reference, and
- keeps the `AF_UNIX` path under the ~104-char `sun_path` limit — pytest's macOS tmp_path (`/private/var/folders/.../pytest-of-runner/...`) overflows it, which is the same root cause behind the other macOS socket failures.

Both fixtures already `rmtree` their `sock_dir` on teardown, so cleanup is unchanged. Verified the modules still collect cleanly.

## Scope / remaining

I kept this to the two **definitely-broken** fixtures (they NameError 100% of the time). The other macOS failures — `test_concurrency`, `test_capture_transcript_no_spawn_defer`, `test_bridge_socket_first` — share the **same `AF_UNIX`-path-too-long root cause** (they build `tmp_path / "sock"`), and pass on Linux today. Happy to route those sockets through short dirs too in a follow-up, but I didn't want to churn currently-green fixtures I can't exercise on a macOS runner locally — your call on whether you'd like that.

### DIFF
--- a/tests/test_bridge_socket_first.py
+++ b/tests/test_bridge_socket_first.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import json
 import os
+import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import time
 from pathlib import Path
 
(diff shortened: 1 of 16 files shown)


CodeAbra:
Thanks @danielhertz1999-bit — the macOS CI fixture/socket fix landed equivalently via #31 (now on main, CI green), so this is covered. Closing as superseded; really appreciate you chasing the red CI.

PR #25, opened 2026-06-27
Fix stale SessionStart cache during active WAKE sessions

Marsu6996:
## Summary
- refresh the SessionStart cache from WAKE/DROWSY, not only after SLEEP
- add single-flight, min-interval and watermark sidecar guards for cheap best-effort refreshes
- make the SessionStart hook TTL default to 12h so stale caches fall through to live recall
- add coverage for refresh decisions, telemetry, sidecar path consistency, TTL fallback and failure cases

## Validation
- .venv/bin/python -m pytest tests/test_session_start_cache_refresh.py tests/test_session_recall_precache.py tests/test_session_refresh_rpc.py tests/test_frozen_constant_hermeticity.py tests/test_consolidation_single_driver.py
- 45 passed in 13.25s

## Notes
- This branch is based on origin/main at v1.2.0.
- It supersedes the earlier branch that was accidentally based on v1.1.7.
- Local production smoke testing showed periodic_wake writes the cache successfully, with session_start_cache_write_started -> session_start_cache_write_success and no failures.

PR #31, opened 2026-06-30
test(socket): use short paths for macOS integration sockets

Marsu6996:
## Summary

Make the socket-heavy integration tests pass reliably on macOS by moving AF_UNIX socket files out of pytest's long per-test temporary directories and into short `tempfile` directories.

This is test infrastructure only. No production daemon, socket, MCP wrapper, or store behavior changes in this PR.

## Root Cause

Darwin's Unix-domain socket path limit is shorter than the paths generated by pytest on macOS runners. In practice, tests were binding paths like:

```text
/private/var/folders/.../pytest-of-runner/.../test_name0/sock/d.sock
```

That can exceed the AF_UNIX `sun_path` limit and fail with either:

- `OSError: AF_UNIX path too long`
- daemon subprocess never binding its socket within the timeout
- wrapper/socket integration tests failing downstream because the fake daemon never became reachable

This was visible in the macOS CI gate and also reproduced locally when running the full suite on macOS.

## What Changed

This PR updates the remaining socket integration fixtures to use short paths created via `tempfile.TemporaryDirectory(prefix="iai-sock-")` or `tempfile.mkdtemp(prefix="iai-sock-")`.

Covered test areas:

- daemon control socket tests
- MCP wrapper bridge tests
- reconnect/death-path tests
- doctor checklist socket checks
- capture-transcript `--no-spawn` defer tests
- socket activity tracking tests
- launchd inherited-fd socket tests
- subagent wrapper reuse tests

It also fixes two fixture bugs that became visible in CI:

- `tests/test_mcp_tools.py::daemon_sock` referenced `tmp_path` from a module-scoped fixture body, where pytest cannot inject it.
- `tests/test_socket_disconnect_reconnect.py::fake_daemon` referenced `tmp_path` without declaring it as a fixture argument.

Finally, the plist rendering test now falls back to stdlib `xml.etree.ElementTree` when `defusedxml` is not installed. The XML being parsed is generated by the test/runtime itself, not untrusted input.

## Relationship To Existing PR #30

This PR intentionally supersedes the partial macOS socket-fixture fix in #30. PR #30 fixes some of the same root cause, but it only touches a subset of the failing files and its macOS CI is still red. This PR is the full-suite-derived version: it includes the remaining socket fixtures that failed later in the suite.

## Validation

Targeted macOS/socket suite:

```bash
.venv/bin/python -m pytest \
  tests/test_bridge_socket_first.py \
  tests/test_capture_transcript_no_spawn_defer.py \
[...]

PR #32, opened 2026-06-30
fix(daemon): refresh SessionStart cache during WAKE

Marsu6996:
## Summary

Refresh the SessionStart precache during WAKE/DROWSY operation, not only after the post-sleep pipeline.

This prevents `~/.iai-mcp/.session-start-payload.cached.md` from staying stale during long-lived active daemon sessions where the store keeps ingesting records but the daemon does not reach SLEEP.

## Root Cause

Before this change, `_write_session_start_cache` was effectively wired to the post-sleep pipeline. That is fine for a daemon that regularly reaches SLEEP, but it is not enough for active Claude/Codex usage where HID activity and MCP traffic can keep the daemon in WAKE/DROWSY for extended periods.

In that regime:

- new records continue to be inserted;
- the live store and runtime graph can move forward;
- the SessionStart precache file can remain old;
- new sessions can receive a stale memory bootstrap.

The bug is not that SessionStart assembly itself is wrong. The bug is that the cached artifact had no cheap active-session refresh path.

## What Changed

Adds a best-effort WAKE-time refresh path:

- `_session_start_cache_watermark(store)` computes a cheap watermark from active records.
- `_should_refresh_session_start_cache(...)` compares the live watermark, sidecar metadata, cache presence, and minimum interval.
- `_maybe_refresh_session_start_cache(...)` wraps refresh attempts in a single-flight lock and degrades by skip reason rather than blocking the daemon.
- `_default_session_start_cache_meta_path(...)` centralizes the sidecar path derivation so writer and reader cannot drift.

Refresh attempts are now triggered from two active-session paths:

- after `pending_embeddings_wake_sequence` when ingest/re-embed moved records;
- during WAKE/DROWSY ticks as a periodic safety net.

The existing post-sleep path still forces a rebuild inside the exclusive sleep pipeline lock, where paying that cost is acceptable.

## Degradation And Telemetry

The active-session refresh is intentionally conservative. It skips instead of doing expensive work when:

- a refresh is already in progress;
- the minimum interval has not elapsed;
- no records changed since the previous watermark;
- the runtime graph cache is cold;
- cache metadata is absent or unreadable;
- the probe fails.

Refresh events are emitted as:

- `session_start_cache_write_started`
- `session_start_cache_write_success`
- `session_start_cache_write_skipped`
- `session_start_cache_write_failed`

Event-write failures are contained so they cannot crash the daemon tick.

## Hook Safety Net

The SessionStart shell hook now has a default-on TTL fallback via:

```text
IAI_MCP_SESSION_CACHE_MAX_AGE_SEC
```

Default behavior refuses to serve very stale precache content indefinitely. Setting the value to `0` preserves the old serve-regardless-of-age behavior explicitly.

## Scope

This PR is intentionally limited to the SessionStart cache refresh bug. It does not include the later recall latency/performance work, the macOS socket test-infra fix, or the Loic-specific journalise capture filtering from the larger experimental branch.

## Validation

Targeted tests:

```bash
.venv/bin/python -m pytest \
  tests/test_session_start_cache_refresh.py \
  tests/test_session_recall_precache.py \
  tests/test_session_refresh_rpc.py \
  -q
```

Result:

```text
35 passed in 12.41s
```

The original monolithic branch that included this change later passed the full local suite after additional split-prep fixes:

```text
3642 passed, 121 skipped, 1 xfailed in 1026.58s
```

## Review Notes

This PR replaces the earlier monolithic PR #25 for the SessionStart-cache portion only. #25 was closed because later commits were pushed to the same branch and made it too broad to review cleanly.

### DIFF
--- a/src/iai_mcp/_deploy/hooks/iai-mcp-session-recall.sh
+++ b/src/iai_mcp/_deploy/hooks/iai-mcp-session-recall.sh
@@ -43,10 +43,23 @@ ts=$(date -u +%Y-%m-%dT%H:%M:%SZ)
 } >> "$log" 2>/dev/null
 
 # precache for SessionStart hook
-# Read the daemon-written cache whenever it is non-empty (no age cap).
+# Read the daemon-written cache whenever it is non-empty AND younger than the
+# max-age safety net. Default: 43200s (12h). Set
+# IAI_MCP_SESSION_CACHE_MAX_AGE_SEC=0 to disable the cap (legacy behaviour:
+# serve the cache regardless of age). Unset / non-numeric → 43200s.
+#
+# Rationale: if the daemon is down, crashed, or its regeneration path is
+# silently failing, we MUST NOT re-inject a multi-day-old prefix into every
+# new Claude/Codex session. The TTL is a back-stop; the primary defence is
+# `daemon._maybe_refresh_session_start_cache`, which refreshes the cache on
+# WAKE whenever new records land.
 # Each branch writes a contract log marker. Falls through to the live CLI path
 # on any miss.
 cache_path="$HOME/.iai-mcp/.session-start-payload.cached.md"
+cache_max_age="${IAI_MCP_SESSION_CACHE_MAX_AGE_SEC-43200}"
+case "$cache_max_age" in
+  ''|*[!0-9]*) cache_max_age=43200 ;;
+esac
 if [ -s "$cache_path" ]; then
   # Cross-platform mtime: try GNU stat, then BSD stat.
   cache_mtime=$(stat -c %Y "$cache_path" 2>/dev/null || stat -f %m "$cache_path" 2>/dev/null || echo 0)
@@ -55,13 +68,17 @@ if [ -s "$cache_path" ]; then
   else
     now_epoch=$(date +%s)
     age=$(( now_epoch - cache_mtime ))
-    cache_out=$(head -c 10000 "$cache_path" 2>/dev/null || true)
-    if [ -n "$cache_out" ]; then
-      printf '%s' "$cache_out"
-      echo "$ts cache-hit age=${age}s bytes=${#cache_out}" >> "$log" 2>/dev/null
-      exit 0
+    if [ "$cache_max_age" -gt 0 ] && [ "$age" -gt "$cache_max_age" ]; then
+      echo "$ts cache-stale age=${age}s max=${cache_max_age}s -> live-cli" >> "$log" 2>/dev/null
+    else
+      cache_out=$(head -c 10000 "$cache_path" 2>/dev/null || true)
+      if [ -n "$cache_out" ]; then
+        printf '%s' "$cache_out"
+        echo "$ts cache-hit age=${age}s bytes=${#cache_out}" >> "$log" 2>/dev/null
+        exit 0
+      fi
+      echo "$ts cache-miss empty (file existed but read returned 0 bytes)" >> "$log" 2>/dev/null
     fi
-    echo "$ts cache-miss empty (file existed but read returned 0 bytes)" >> "$log" 2>/dev/null
   fi
 elif [ -e "$cache_path" ]; then
   echo "$ts cache-miss empty (zero-byte file)" >> "$log" 2>/dev/null
--- a/src/iai_mcp/daemon/__init__.py
+++ b/src/iai_mcp/daemon/__init__.py
@@ -71,7 +71,16 @@ S4_FIRST_ITER_GRACE_SEC: float = float(
 )
 
 SESSION_START_CACHE_PATH = Path.home() / ".iai-mcp" / ".session-start-payload.cached.md"
+SESSION_START_CACHE_META_PATH = (
+    Path.home() / ".iai-mcp" / ".session-start-payload.cached.meta.json"
+)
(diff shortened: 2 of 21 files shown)


Marsu6996:
This PR currently depends on #31 for macOS CI stability. The failing run here is the pre-#31 socket-path issue on Darwin (`AF_UNIX path too long`, `socket never bound`, and the fixture `tmp_path` errors), not a failure introduced by the WAKE SessionStart cache refresh itself.

Once #31 is merged into `main`, I will rebase/refresh this branch and rerun CI cleanly against the fixed test infrastructure.

danielhertz1999-bit:
The macOS CI failure here (`NameError: tmp_path`, `OSError: AF_UNIX path too long`, daemon socket-accept timeouts in `test_mcp_tools.py`, `test_bridge_socket_first.py`, `test_concurrency.py`, `test_daemon_crash_loop_immunity.py`, etc.) is pre-existing and unrelated to this PR's diff (`iai-mcp-session-recall.sh` / `daemon/__init__.py`) — it's the same root cause already fixed comprehensively in #31. I merged #31 into my own #27/#28/#29/#30 to unblock their CI; you may want to do the same here (or just rebase once #31 lands).

Marsu6996:
Following @danielhertz1999-bit's suggestion above, I've merged `pr/macos-socket-test-infra` (#31) into this branch as a real merge commit (no squash, no cherry-pick) so the macOS CI here can validate the WAKE SessionStart cache-refresh changes against the fixed test infrastructure.

**For reviewers**: please only assess the WAKE diff (`iai-mcp-session-recall.sh` and `daemon/__init__.py`). The 16 test files now visible in the diff belong to #31.

Once #31 lands on `main`, I'll refresh this branch to drop the merged-in commits and restore a clean diff. If #31 ends up squash-merged rather than merge-committed, I'll merge `main` back into this branch to resync.

Marsu6996:
Thanks for the heads-up, @danielhertz1999-bit — appreciated. CI now green here after merging #31 into this branch. I'll refresh once #31 lands on `main`.

danielhertz1999-bit:
Did a close read of the daemon logic (draft, so treat this as WIP feedback, not a merge gate). The design is genuinely nice — the single-SELECT watermark, the `.meta.json` sidecar, the non-blocking single-flight lock, and especially the `force_rebuild` split (SLEEP holds EX so it can absorb the runtime-graph rebuild; WAKE skips with `runtime_graph_cache_cold`) all hang together well. Using `max_vec_label` as the primary signal is the right call — it catches backfilled records an `MAX(created_at)` comparison would miss. Test isolation via the `conftest` meta-path redirect is also exactly right.

Four things worth considering before it comes out of draft:

**1. (medium) Event-log churn on the periodic path.** `_maybe_refresh_session_start_cache(trigger="periodic_wake")` runs on essentially every WAKE/DROWSY lifecycle tick (~30s), and on a quiet daemon it emits a `session_start_cache_write_skipped` event every time (`no_new_records` / `min_interval_not_elapsed`). That's ~2,880 info events/day with nothing changing. The probe itself is cheap as documented, but the steady-state skip-event write will grow the events table. Might be worth suppressing the no-op skips (emit only on a reason *transition*, or drop the common ones to `log.debug`).

**2. (low-med) The authoritative SLEEP write can be silently skipped.** The single-flight lock is priority-agnostic + non-blocking, so `trigger=sleep_pipeline, force_rebuild=True` returns `refresh_in_progress` if a `periodic_wake` refresh happens to hold the lock at that moment. The cache won't be wrong (WAKE only writes when the graph is warm), but the rebuild the SLEEP path intended gets dropped for that cycle, where previously the SLEEP write always ran. Might want SLEEP to block-acquire (or otherwise take priority) rather than skip.

**3. (low) Failing watermark probe → refresh churn.** When the probe hits its `except` and returns the `-1` sentinel dict, that never matches the stored watermark, so `_should_refresh` returns `watermark_changed` and a refresh is attempted every interval. It's bounded by the min-interval gate and the write is guarded, so it won't crash — but a persistent schema/probe failure would retry indefinitely. Could treat a `-1` probe as `skip(reason=probe_failed)` instead of a change signal.

**4. (info / coordination) This stacks on #31.** The 16 socket-fixture files here are #31's changes folded in, which is what makes CI green. Assuming #31 merges first, this rebases down to just the daemon + hook + tests. Flagging so the merge order is intentional.

Really solid work on a real problem (the stale-precache-for-long-lived-WAKE case). Happy to dig into any of these further.

CodeAbra:
Merged into main (squash) — thank you, @Marsu6996! SessionStart cache now refreshes during WAKE.

Marsu6996:
Thanks again for the detailed review, @danielhertz1999-bit.

CodeAbra squash-merged both #31 and this PR into `main`. I incorporated your feedback before the squash:

- #31 now drops the `xml.etree` fallback and imports `defusedxml.ElementTree` directly, since `defusedxml>=0.7` is declared for that plist lint path.
- This PR now suppresses the common no-op `periodic_wake` skip events for `min_interval_not_elapsed` / `no_new_records`, while keeping real skip/failure reasons observable.
- Watermark probe failures now skip with `probe_failed` instead of treating the `-1` sentinel as a changed watermark.
- The authoritative `sleep_pipeline` refresh now waits for an in-flight refresh, with a bounded timeout so it degrades back to an observable `refresh_in_progress` skip rather than risking an unbounded daemon hang.

Appreciate the close read; it made the final squash cleaner.

PR #33, opened 2026-06-30
doctor: Windows-safe pid liveness in check (a) and heartbeat scanner

danielhertz1999-bit:
## Problem

Two `os.kill(pid, 0)` liveness probes were left unguarded after the Windows port. On Windows `os.kill` rejects signal 0 with `OSError [WinError 87] (The parameter is incorrect)`, so on a perfectly healthy daemon `iai-mcp doctor` reports:

```
[FAIL] (a) daemon process alive    liveness probe failed: OSError: [WinError 87] The parameter is incorrect
[FAIL] (m) heartbeat scanner       could not scan ...\wrappers: OSError: [WinError 87] The parameter is incorrect
```

- **check (a)** — `os.kill(pid, 0)` raises before the psutil refinement runs.
- **(m) heartbeat scanner** — `_is_pid_alive` raises *inside* `HeartbeatScanner.scan()` (which only narrows `OSError` around the glob, not the per-pid probe), so it escapes as `could not scan wrappers`.

`check_b` was already given the Windows TCP-transport treatment; this brings (a) and the heartbeat scanner in line.

## Fix

Mirror the platform split already used by `lifecycle_lock._is_pid_alive`:

- **Windows** → use `psutil` (`pid_exists` for the wrapper probe; `Process().cmdline()` for the daemon-identity check).
- **POSIX** → unchanged `os.kill(pid, 0)`.

check (a) also handles `psutil.NoSuchProcess` explicitly so a dead pid reads `no process found` rather than the generic `could not verify`.

## Verification

- On Windows the live daemon now reports `[PASS] (a) ... PID <n> (iai_mcp.daemon)` and `[PASS] (m) heartbeat scanner n=… fresh`.
- POSIX behaviour is unchanged (guards are Windows-only). `tests/test_doctor_checklist.py` and `tests/test_heartbeat_scanner.py` pass (24 passed, 1 skipped).

### DIFF
--- a/src/iai_mcp/doctor/_lifecycle_checks.py
+++ b/src/iai_mcp/doctor/_lifecycle_checks.py
@@ -65,44 +65,66 @@ def check_a_daemon_alive() -> CheckResult:
             f"daemon_pid={pid!r} is not a valid PID (corrupt state?)",
         )
 
+    # os.kill(pid, 0) is the POSIX liveness idiom, but on Windows os.kill
+    # rejects signal 0 with OSError [WinError 87] (invalid parameter), so a
+    # perfectly healthy daemon reports "liveness probe failed". Skip the probe
+    # there and rely on the psutil refinement below, which both confirms the
+    # pid exists and that it is actually an iai_mcp.daemon. Mirrors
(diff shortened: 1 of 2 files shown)


danielhertz1999-bit:

PR #34, opened 2026-06-30
tests: route remaining daemon sockets through short mkdtemp dirs (macOS CI)

danielhertz1999-bit:
## Problem

Four socket fixtures still build their AF_UNIX path under pytest's `tmp_path` (`sock_dir = tmp_path / "sock"`). On macOS that path lives under `/private/var/folders/.../pytest-of-runner/pytest-N/<test-name>0/`, which — once `/sock/d.sock` is appended — can exceed the ~104-char `sun_path` limit and raise `OSError: AF_UNIX path too long`.

These are the files **not** covered by #31 (which fixes the same root cause across the other ~14 socket fixtures):

- `tests/test_capture_transcript_no_spawn.py`
- `tests/test_concurrent_wrapper_spawn.py`
- `tests/test_daemon.py`
- `tests/test_doctor_apply_recovery.py`

This PR is complementary to #31 — disjoint file set, no overlap.

## Fix

Route each through `tempfile.mkdtemp(prefix="iai-sock-")`, matching the pattern already used elsewhere. Fixtures that `rmdir`'d `sock_dir` in teardown continue to do so (now removing the mkdtemp dir); the daemon-down helpers never bind the socket, so only an empty temp dir is created and the OS reclaims it.

## Scope

POSIX-only test paths; no production code touched. Files compile and collect cleanly. Runtime behaviour is verified by the macOS CI job (the platform the bug manifests on).

### DIFF
--- a/tests/test_capture_transcript_no_spawn.py
+++ b/tests/test_capture_transcript_no_spawn.py
@@ -5,6 +5,7 @@ import os
 import platform
 import subprocess
 import sys
+import tempfile
 import time
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


danielhertz1999-bit:

PR #35, opened 2026-06-30
doctor: WARN checks no longer count as FAIL or drive the exit code

danielhertz1999-bit:
## Problem

`CheckResult` distinguishes `status == "WARN"` (advisory) from `FAIL`, and `print_checklist` renders them as `[WARN]`. But the exit-code tally counted them as failures:

```python
fail_count = sum(1 for r in results if not r.passed)   # WARNs have passed=False
```

So an advisory like (x) collapsed-timestamps produced self-contradicting output:

```
[WARN] (x) no collapsed-timestamp groups   ... run 'iai-mcp migrate --rederive-timestamps' to repair
1/25 FAIL. Run with --apply to attempt recovery. Exit 1.
```

— "1/25 FAIL ... Exit 1" with **zero `[FAIL]` rows on screen**, and a non-zero exit for a purely advisory condition that has no repair action (so `--apply` then reports `Exit 2` "still broken" against something it can never fix).

## Fix

Classify a hard fail as `not r.passed and r.status != "WARN"` in both the summary tally and the post-`--apply` recheck. When only WARNs remain:

```
All checks passed (1 warning(s)). Exit 0.
```

WARNs still render as `[WARN]`; no check definitions change.

## Note for maintainers

This **changes exit-code semantics**: `doctor` previously returned 1 whenever any WARN was present (e.g. collapsed timestamps, no HID idle source, CLI not on PATH); it now returns 0 when all *hard* checks pass. That matches the `[WARN]` rendering and the advisory wording, but if you intend a specific WARN to gate CI, the alternative is to promote it to a real `FAIL`. Happy to adjust.

## Verification

- New test `test_warn_does_not_count_as_fail_or_drive_exit`: an only-WARN result yields Exit 0, "1 warning", and no `FAIL` summary line.
- `tests/test_doctor_checklist.py`: 14 passed, 1 skipped.

### DIFF
--- a/src/iai_mcp/doctor/__init__.py
+++ b/src/iai_mcp/doctor/__init__.py
@@ -553,10 +553,18 @@ def cmd_doctor(args: argparse.Namespace) -> int:
         print(hint)
         print()
     print_checklist(results)
-    fail_count = sum(1 for r in results if not r.passed)
+    # A WARN is advisory: it renders as [WARN] and must NOT count as a FAIL or
+    # drive the exit code (otherwise e.g. the (x) collapsed-timestamp advisory
+    # prints "1/25 FAIL ... Exit 1" with zero [FAIL] rows on screen). Only a
+    # genuine non-passing, non-WARN check is a hard fail.
+    fail_count = sum(1 for r in results if not r.passed and r.status != "WARN")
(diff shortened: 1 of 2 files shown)


danielhertz1999-bit:

PR #36, opened 2026-06-30
daemon: auto-clear expired sleep quarantine at boot

danielhertz1999-bit:
## Problem

An expired sleep-cycle quarantine is only cleared when a sleep cycle is *next attempted* (`SleepPipeline.run` → `_check_and_maybe_auto_recover_quarantine`, line ~397). A daemon that never reaches a sleep window — e.g. pinned in WAKE with active sessions — therefore carries a zombie quarantine entry in `lifecycle_state.json` indefinitely.

Observed: a daemon still reporting

```
[WARN] (l) sleep cycle quarantine   quarantine expired (until=2026-06-27T...); reason=sleep step 3 (DREAM_DECAY) failed 3x
```

three days after the TTL elapsed, because no sleep cycle had run in the interim. `is_quarantined()` correctly returns `False` for it, but never tidies the stale dict.

## Fix

Add `SleepPipeline.recover_expired_quarantine()` — a thin public wrapper over the existing `_check_and_maybe_auto_recover_quarantine()` that returns whether it cleared an expired/malformed entry — and call it **once at daemon boot**, right after the pipeline is constructed (alongside the existing lifecycle boot-normalization). Failures are swallowed so boot can never wedge on it.

Active quarantines and the sleep-cycle entry path are unchanged.

## Verification

New tests in `test_sleep_pipeline.py`:
- `test_recover_expired_quarantine_clears_without_sleep_cycle` — expired entry cleared without running a cycle.
- `test_recover_expired_quarantine_keeps_active_quarantine` — active entry survives.
- `test_recover_expired_quarantine_noop_when_absent` — no-op when none present.

Full file: 29 passed.

### DIFF
--- a/src/iai_mcp/daemon/__init__.py
+++ b/src/iai_mcp/daemon/__init__.py
@@ -1147,6 +1147,20 @@ async def main() -> int:
         _idle_detector = _IdleDetector()
         _sleep_pipeline = _SleepPipeline(store=store)
 
+        # Eagerly clear a long-expired quarantine at boot. Auto-recovery
+        # otherwise only fires when a sleep cycle is next attempted, so a daemon
+        # that never reaches a sleep window (e.g. pinned in WAKE with active
+        # sessions) would carry a zombie quarantine entry indefinitely and keep
+        # `doctor` (l) reporting a quarantine that expired days ago.
+        try:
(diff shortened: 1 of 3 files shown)


danielhertz1999-bit:

PR #38, opened 2026-07-01
fix(daemon): defer startup drains behind MCP activity

Marsu6996:
## Problem

The daemon still had too much work on its startup hot path. On a store with deferred captures or capture-queue backlog, startup could begin draining work before the MCP socket was ready and before the daemon knew whether a client was actively trying to connect.

That is risky for MCP clients because the process can look started while the useful interface is not yet responsive. In practice, the expensive startup work competes with first MCP traffic, which can turn a normal client launch into a slow or disconnected handshake.

There was a second related hot-path issue: WAKE / boot SessionStart refresh could force a full runtime-graph rebuild. That is useful work, but it is the wrong moment to pay for it if the daemon is trying to become responsive.

## Proposed solution

This PR narrows startup to the minimum needed to serve MCP promptly:

- start serving the MCP socket before running capture-queue or deferred-capture drains;
- only run startup drains once MCP traffic has been quiet long enough;
- persist retry markers when startup drains are skipped or fail, so the scheduler can retry later instead of silently dropping work;
- keep WAKE / boot SessionStart refresh on the structural runtime-graph cache instead of forcing a full runtime graph rebuild on the hot path.

The important trade-off is intentional: startup responsiveness wins over immediate background cleanup. The cleanup is still durable and retried; it is just moved out of the client-facing startup window.

## Safety and behavior

On a genuinely cold boot with no structural overlay or snapshot, `load_recall_structural()` can return the `cold_degrade` empty assignment instead of building the graph immediately. That is deliberate here. The first recall or a later sleep/drowsy path can pay the rebuild cost, while daemon startup and MCP serving stay responsive.

Startup drain failures and skips remain observable through events such as `capture_queue_drain_failed`, `capture_queue_drain_startup_skipped`, and `deferred_drain_startup_skipped`.

## Scope

This is rebased on `v1.2.1` (`88a0f0d`). `v1.2.1` already includes the macOS socket CI fix and the WAKE SessionStart refresh from the earlier PRs, so this PR is intentionally limited to the remaining daemon startup responsiveness work.

## Verification

- `python -m py_compile src/iai_mcp/daemon/__init__.py`
- `git diff --check`
- `pytest tests/test_daemon_tick_yields_to_mcp_activity.py tests/test_session_start_cache_refresh.py -q`
  - `33 passed`
- `pytest tests/test_capture_queue.py tests/test_socket_server_dispatch.py tests/test_daemon_crash_loop_immunity.py -q`
  - `31 passed`
- `env -u IAI_RECALL_READ_TIMEOUT pytest -m "not perf and not slow and not live" --ignore=tests/test_embedder_numeric_parity.py`
  - `3612 passed, 32 skipped, 87 deselected, 1 xfailed`
- `env -u IAI_RECALL_READ_TIMEOUT pytest tests/test_embedder_numeric_parity.py -q -s`
  - `4 passed, 1 skipped`

## Cross-check

Reviewed with Claude before publishing. The review agreed with keeping startup responsive, preserving deferred work through retry markers, and avoiding forced graph rebuilds on the daemon hot path.

## Attribution

### DIFF
--- a/src/iai_mcp/daemon/__init__.py
+++ b/src/iai_mcp/daemon/__init__.py
@@ -61,6 +61,12 @@ VALID_TRANSITIONS: dict[str, set[str]] = {
 }
 
 TICK_INTERVAL_SEC: int = 30
+# Read once at import so spawned daemon processes can override it via env.
+STARTUP_DEFERRED_DRAIN_GRACE_SEC: float = float(
+    os.environ.get("IAI_MCP_STARTUP_DEFERRED_DRAIN_GRACE_SEC", "20.0")
+)
+CAPTURE_QUEUE_RETRY_PENDING_KEY: str = "_capture_queue_drain_retry_pending"
+CAPTURE_QUEUE_LAST_RETRY_KEY: str = "_capture_queue_drain_last_retry_at"
 
 DEFAULT_CYCLE_COUNT: int = 4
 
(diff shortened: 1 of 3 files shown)


CodeAbra:
Merged into main (squash) — thanks @Marsu6996! Startup drains now defer behind MCP readiness.

PR #39, opened 2026-07-01
fix(store): preserve structural metadata in SQL paths

Marsu6996:
## Problem

Several SQL-backed read and maintenance paths were not preserving the full structural metadata that newer recall paths rely on.

The concrete failure mode is subtle: records can carry structural fields such as `structure_hv`, `hv_tier`, and `structure_hv_payload`, and they can also be marked `embedding_pending` while a safe re-embed or migration is still in progress. Some explicit SQL projections returned only the older semantic fields, so structural recall and daemon-down fallback paths could see incomplete records.

There was also a write-path risk: normal `MemoryStore.insert()` row writes could lose the caller's `embedding_pending` value by falling back to the database default. During reembed staging, a placeholder target-dimension embedding can exist before the row is actually ready; if `embedding_pending` is lost, downstream readers may treat the row as ready too early.

## Proposed solution

This PR makes the SQL paths preserve the same structural metadata contract as the higher-level store APIs:

- include `structure_hv` in explicit SQL projections used by `get_batch`, `recent_pending_markers`, and direct recency / daemon-down reads;
- preserve `embedding_pending` on normal row writes instead of relying on the database default;
- carry `hv_tier`, `structure_hv_payload`, and `embedding_pending` through reembed staging.

The goal is not to change recall ranking. It is to make every read path see the same record shape, especially during degraded or migration paths where these SQL projections are used directly.

## Safety and behavior

Pending rows remain pending during reembed migration even if the migration writes a target-dimension placeholder embedding. The normal pending drain remains responsible for flipping readiness once the real embedding is available.

This keeps the crash-safe migration behavior intact: staging can make progress without accidentally advertising incomplete rows as ready.

## Scope

Based on `v1.2.1` / current `origin/main` (`88a0f0d`). This is independent of the daemon startup and socket-concurrency PRs.

## Verification

- `.venv/bin/python -m py_compile src/iai_mcp/store/_store.py src/iai_mcp/hippo/_recall.py src/iai_mcp/migrate/_reembed.py`
- `git diff --check`
- `.venv/bin/python -m pytest tests/test_store.py::test_sql_record_projection_preserves_structure_and_pending tests/test_store.py::test_direct_recency_projection_preserves_structure_metadata tests/test_store.py::test_recent_pending_markers_preserve_structure_metadata tests/test_migrate_reembed_crash_safe.py::test_migration_preserves_structural_metadata_and_pending_flag -q`
- `.venv/bin/python -m pytest tests/test_store.py tests/test_store_get_batch.py tests/test_migrate_reembed_crash_safe.py -q`
- `.venv/bin/python -m pytest tests/test_semantic_degraded_path.py tests/test_schema_v2.py tests/test_tem_store.py tests/test_direct_recency_read.py -q`
- `env -u IAI_RECALL_READ_TIMEOUT .venv/bin/python -m pytest -m "not perf and not slow and not live" --ignore=tests/test_embedder_numeric_parity.py`
  - `3611 passed, 32 skipped, 87 deselected, 1 xfailed`
- `env -u IAI_RECALL_READ_TIMEOUT .venv/bin/python -m pytest tests/test_embedder_numeric_parity.py -q -s`
  - `4 passed, 1 skipped`

## Cross-check

Reviewed with Claude before publishing. Claude approved this as a standalone PR and specifically asked that the PR description call out the write-path `embedding_pending` fix, because that is the part most likely to be missed in a metadata-only looking patch.

## Attribution

### DIFF
--- a/src/iai_mcp/hippo/_recall.py
+++ b/src/iai_mcp/hippo/_recall.py
@@ -17,7 +17,7 @@ _DIRECT_RECENCY_SQL = (
     " stability, difficulty, last_reviewed, never_decay, never_merge,"
     " provenance_json, created_at, updated_at, tags_json, language,"
     " s5_trust_score, profile_modulation_gain_json, schema_version,"
-    " hv_tier, structure_hv_payload,"
+    " structure_hv, hv_tier, structure_hv_payload,"
     " COALESCE(embedding_pending, 0) AS embedding_pending"
     " FROM records WHERE tombstoned_at IS NULL ORDER BY created_at DESC"
 )
@@ -29,7 +29,7 @@ _DIRECT_RECENCY_SQL_LIMITED = (
     " stability, difficulty, last_reviewed, never_decay, never_merge,"
     " provenance_json, created_at, updated_at, tags_json, language,"
     " s5_trust_score, profile_modulation_gain_json, schema_version,"
-    " hv_tier, structure_hv_payload,"
+    " structure_hv, hv_tier, structure_hv_payload,"
     " COALESCE(embedding_pending, 0) AS embedding_pending"
     " FROM records WHERE tombstoned_at IS NULL ORDER BY created_at DESC"
     " LIMIT ?"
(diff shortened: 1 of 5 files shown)


CodeAbra:
Merged into main (squash) — thanks @Marsu6996! SQL paths now carry the full structural metadata.

PR #40, opened 2026-07-01
fix(socket): bound concurrent memory recall dispatch

Marsu6996:
## Problem

`memory_recall` is the most expensive MCP request path, and the socket daemon previously treated concurrent cued recalls like any other JSON-RPC method: every request could immediately enter `dispatch()` in a worker thread.

That is fine for isolated calls, but it becomes fragile during bursty client behavior. A few simultaneous recall requests can pile up expensive recall work, compete for CPU, and make the daemon slower at serving the very client traffic that should keep it responsive. In the worst case, overload looks like a hang rather than an honest degraded response.

There was also a smaller fail-fast issue on the CLI side: socket cleanup called `writer.wait_closed()` without a bound. If the peer or platform got stuck during close, a path that was supposed to fail quickly could still wait too long.

## Proposed solution

This PR adds a narrow backpressure layer only around cued `memory_recall` dispatches:

- allow a small number of cued recalls to run concurrently (`IAI_MCP_RECALL_CONCURRENCY`, default `2`);
- wait briefly for a recall slot (`IAI_MCP_RECALL_SLOT_WAIT_SEC`, default `0.25`);
- if no slot is available, return an explicit degraded empty recall with `_reason: recall_busy`;
- keep non-recall methods, and malformed / cue-less recall validation, on the existing dispatch path;
- bound CLI socket `wait_closed()` cleanup to keep fail-fast behavior actually bounded.

The intent is to make overload explicit and cheap instead of letting recall work queue unboundedly behind the socket server.

## Safety and behavior

A burst of more than two cued recalls within the slot wait window can now receive:

```json
{"hits": [], "_degraded": true, "_reason": "recall_busy"}
```

That is intentional. Normal single-client recall remains unchanged, while overload degrades honestly as an empty recall instead of silently accumulating background work.

The response shape follows the existing degraded recall pattern (`hits: []`, `_degraded: true`, `_reason: ...`), so current clients continue to handle it like a no-match / degraded recall. The new test also verifies that the overflow request never reaches `dispatch()`, which is the important backpressure invariant.

## Scope

This is independent of #38 and #39. It does not include the older personal/noise-filtering branch work, nor the old core/store/top_k changes from the historical branch. Those were either already upstreamed, split into earlier PRs, or intentionally left out.

Only source, tests, and README documentation for the new operator knobs are included.

## Verification

- `.venv/bin/python -m py_compile src/iai_mcp/socket_server.py src/iai_mcp/cli/__init__.py`
- `git diff --check`
- `.venv/bin/python -m pytest tests/test_socket_server_dispatch.py tests/test_iai_recall_fail_fast.py -q`
  - `15 passed in 2.90s`
- `env -u IAI_RECALL_READ_TIMEOUT .venv/bin/python -m pytest tests/test_socket_server_dispatch.py tests/test_iai_recall_fail_fast.py tests/test_socket_activity_tracking.py tests/test_socket_disconnect_reconnect.py tests/test_socket_fail_loud.py tests/test_socket_first_store_hermeticity.py tests/test_socket_inherit_launchd_fd.py tests/test_socket_subagent_reuse.py tests/test_cli_daemon.py tests/test_cli_subprocess_daemon_down.py tests/test_client_surfaces_daemon_down.py tests/test_recall_daemon_down_degrade.py tests/test_mcp_tools.py tests/test_daemon_dispatcher.py -q`
  - `103 passed, 6 warnings in 20.89s`
- `env -u IAI_RECALL_READ_TIMEOUT .venv/bin/python -m pytest -m "not perf and not slow and not live" --ignore=tests/test_embedder_numeric_parity.py -q`
  - `3609 passed, 32 skipped, 87 deselected, 1 xfailed, 375 warnings in 840.54s (0:14:00)`
- `env -u IAI_RECALL_READ_TIMEOUT .venv/bin/python -m pytest tests/test_embedder_numeric_parity.py -q -s`
  - `4 passed, 1 skipped in 284.99s (0:04:44)`

## Cross-check

Reviewed with Claude before publishing. Claude's verdict was: correct, well-scoped, low-regression, ship as a standalone PR. Claude specifically checked the cue-less validation path, semaphore lifecycle, degraded response shape, Python 3.11 semaphore behavior, env guards, and the `IAI_RECALL_READ_TIMEOUT` hermeticity guard.

## Upstream check before publish

Immediately before opening this PR:

- latest release was still `v1.2.1` (`2026-07-01T06:03:21Z`)
- `origin/main` was `88a0f0d Release v1.2.1`
- open public PRs were only #38 and #39
- #38 and #39 were mergeable, green, and had no comments or reviews

## Attribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,8 @@ Measured on an Apple M2 Max (64 GB). The harnesses are the proof — run them yo
 |---|---|---|
 | `IAI_MCP_STORE` | `~/.iai-mcp/` | Data directory |
 | `IAI_MCP_PYTHON` | — | Absolute path to the venv Python (for the MCP host config) |
+| `IAI_MCP_RECALL_CONCURRENCY` | `2` | Maximum cued `memory_recall` calls dispatched concurrently by the socket daemon |
+| `IAI_MCP_RECALL_SLOT_WAIT_SEC` | `0.25` | How long an overflow cued recall waits for a slot before returning `_degraded: recall_busy` |
 
 The old `IAI_MCP_EMBED_MODEL` knob is gone — the embedder is a single built-in English-only model. There are many internal tuning knobs (`IAI_MCP_*`), but you shouldn't need to touch them.
 
--- a/src/iai_mcp/cli/__init__.py
+++ b/src/iai_mcp/cli/__init__.py
@@ -130,8 +130,8 @@ def _send_jsonrpc_request(
         finally:
             try:
                 writer.close()
-                await writer.wait_closed()
-            except OSError:
+                await asyncio.wait_for(writer.wait_closed(), timeout=0.25)
+            except (OSError, asyncio.TimeoutError):
                 pass
 
     try:
@@ -162,8 +162,8 @@ def _send_socket_request(req: dict, *, timeout: float = 30.0) -> dict | None:
         finally:
             try:
                 writer.close()
-                await writer.wait_closed()
-            except OSError:
+                await asyncio.wait_for(writer.wait_closed(), timeout=0.25)
+            except (OSError, asyncio.TimeoutError):
                 pass
 
     return asyncio.run(_runner())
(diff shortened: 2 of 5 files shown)


CodeAbra:
Merged into main (squash) — thank you @Marsu6996! Concurrent recall dispatch is now bounded; documented the two env knobs in the README.

PR #41, opened 2026-07-01
profile: let MCP clients set bool/numeric knobs (undo JSON-type loss at the edge)

danielhertz1999-bit:
## Problem

Setting a **bool** (or numeric) profile knob through the `profile_get_set` MCP tool fails:

```
profile_set  knob=inertia_awareness  value=true
-> {"status":"error","reason":"value must be bool, got str","schema":"bool"}
```

Some JSON-RPC / MCP clients don't preserve the JSON type of a tool argument — the boolean `true` reaches the daemon as the **string** `"true"`. `profile_set`'s validator is (intentionally) strict and type-exact, so it rejects it. Net effect: every bool / `*_range` knob is effectively **unsettable** through those clients. The only workaround is to bypass the tool entirely and send a hand-crafted request to the daemon socket with a real Python bool — which is how `inertia_awareness` had to be set in practice.

## Fix — at the edge, not in the strict core

The strict typing is deliberate (it's locked by `test_profile_set_live_bool_rejects_non_bool` and `test_profile_set_scene_construction_scaffold_rejects_string`), so this does **not** touch `profile_set`.

Instead it adds `coerce_json_stringified(schema, value)` and calls it in the **core dispatch** for `method=profile_set`, before delegating. It reverses *only* the exact JSON stringification:

- `"true"` / `"false"` (case-insensitive) → the bool
- a numeric string → `int` / `float` for `int_range` / `float_range` knobs
- `dict` values coerced element-wise against the inner schema

Loose spellings that `profile_set` rejects **on purpose** — the int `1`, `"yes"`, `"1"`, `"on"` — are deliberately left untouched, so the strict contract and its existing tests are unchanged. `enum` knobs are already string-typed and pass through.

## Tests

New unit + end-to-end tests in `tests/lilli/test_profile.py` (coercion of `"true"/"false"`, numeric strings, dict recursion; explicit assertions that `1` / `"yes"` are NOT widened; and a coerce-then-`profile_set` round-trip). Full file: **31 passed**, including the pre-existing strict-rejection tests.

## Related (not in this PR)

This fixes the *mechanism* for writing self-report knobs. The larger gap — that sleep consolidation doesn't derive self-report knobs (`dunn_quadrant`, `inertia_awareness`) from an *answered* curiosity question in episode content, so those answers never apply without a manual `profile_set` — is a separate, larger design change worth its own discussion.


CodeAbra:
Merged into main (squash) — thanks @danielhertz1999-bit! String-encoded bool/numeric knobs are now coerced at the edge. Closed rather than shown as merged because it landed via squash.

PR #42, opened 2026-07-01
profile: persist explicitly-set knobs across restarts (durable pinned overrides)

danielhertz1999-bit:
## Problem

`_profile_state` (`core/__init__.py`) is initialised once to `default_state()` and **never rehydrated**. Every daemon restart silently reverts *all* profile knobs to defaults. So a user can answer a curiosity question and set e.g. `dunn_quadrant` / `inertia_awareness` via `profile_set` — and it vanishes on the next restart. The sleep digest then re-asks the same question indefinitely (observed: the same profile question re-fired across 5 consecutive nightly digests despite being answered).

`profile_set` writes a `profile_updated` audit event, but nothing ever replays it.

## Fix — durable *pinned* overrides

Persist the knobs a user **explicitly** set, and reload them at boot:

- **`knobs.py`**: `PROFILE_OVERRIDES_PATH` (`~/.iai-mcp/.profile-knobs.json`) + `save_profile_overrides()` (atomic write, best-effort — never fails the set it's attached to) and `load_profile_overrides()` (**re-validates every entry** against the current schema; drops unknown / deferred / now-invalid knobs so a stale or hand-edited file can't inject bad state).
- **`core`**: track `_pinned_knobs`. The `profile_set` dispatch pins + persists on an accepted set; `rehydrate_profile_overrides()` replays them into `_profile_state` at boot. A pinned knob is **skipped by the `profile_update_from_signal` (bayesian) path**, so an explicit answer *outranks* an inferred one instead of drifting.
- **`daemon`**: call `rehydrate_profile_overrides()` right after the store opens at boot (best-effort; never blocks boot).
- **`conftest`**: redirect `PROFILE_OVERRIDES_PATH` in the hermetic-paths fixture so no test writes the developer's real `~/.iai-mcp/.profile-knobs.json`.

## Scope / design notes

- **Explicit sets only.** Bayesian tuning mutates `_profile_state` directly (not through the `profile_set` dispatch), so learned values are intentionally *not* persisted — only user-stated ones. This is what makes "explicit answer wins" clean.
- `profile_set`'s strict validation is untouched; persistence lives at the dispatch edge.
- Independent of the set-path coercion fix (#41) — they don't overlap.

## Tests

New `tests/lilli/test_profile_durability.py`: save/load round-trip, re-validation drop (unknown / invalid enum / out-of-range), boot rehydrate + pin, dispatch persist + pin, rejected set is neither pinned nor persisted, and a pinned knob survives a contrary bayesian signal. Full `lilli` suite: **348 passed, 1 skipped**.


CodeAbra:
Merged into main (squash) — thank you @danielhertz1999-bit! Pinned profile overrides now survive restarts (rehydrated on startup), so answered questions stop re-firing. Combined cleanly with #41's coercion.

PR #44, opened 2026-07-04
Windows: fix daemon liveness, crypto key corruption, and auth-token resilience

danielhertz1999-bit:
A set of Windows daemon fixes discovered while running the memory engine on
Windows 11. Each is a POSIX-idiom-on-Windows bug that misbehaves silently.
Branch is `main` + these commits only.

## 1. `os.kill(pid, 0)` liveness probes fail on Windows (WinError 87)

`os.kill(pid, 0)` is the POSIX "is this PID alive?" idiom, but on Windows
`os.kill` rejects signal `0` with `OSError [WinError 87]` **even for a live
process**. Three call sites treated that as "process dead", making a healthy,
ticking daemon report as down:

- `doctor` check (a) "daemon process alive" → live daemon reported dead
- `doctor` check (m) heartbeat scanner → aborted with WinError 87
- maintenance pre-flight guard → *always* concluded "no daemon running" on
  Windows, which could let maintenance run against a store the live daemon holds

Each now uses `psutil.pid_exists` (cross-platform; psutil is already a hard
dep), falling back to `os.kill` only on POSIX — matching the existing pattern
in `capture.py` and `lifecycle_lock._is_pid_alive`.

## 2. Auth-token file is a single point of failure with no self-heal

On Windows the daemon authenticates clients with a token it writes to
`~/.iai-mcp/.daemon.token` at startup and holds in memory. If that file later
vanishes out-of-band (external cleanup, AV, a stale-file sweep) while the daemon
runs, the TCP port stays up but **every new client fails the auth handshake** —
old connections survive, so the daemon looks alive while new sessions and
`daemon status` all report "not running". Observed in production.

Adds `_ipc.reassert_token_if_missing()` (idempotent; no-op on POSIX / in client
processes / when the file exists), called as the first daemon tick step, so a
vanished token self-heals within one tick instead of needing a manual restart.

`doctor` check (b) also now reports this precisely — the real endpoint
(`127.0.0.1:<port>`) and "auth token file missing (restart to regenerate)"
instead of the old misleading "`.daemon.sock` present but unreachable".

## 3. Crypto AES key corruption on Windows (missing `O_BINARY`)

`crypto._try_file_set` used `os.open` without `os.O_BINARY`; on Windows that
defaults to text mode and translated `0x0A` bytes in the raw AES key to
`0x0D0A`, silently corrupting it. Adds `os.O_BINARY` (no-op on POSIX).

## Tests

Test-portability changes so the suite runs green on Windows, plus a determinism
fix for a pre-existing centrality-bench flake (unseeded `uuid4()` let near-tied
top-K scores flip a `str(id)` tie-break between runs). New unit tests cover the
token self-heal. No behavior change on POSIX.

### DIFF
--- a/bench/contradiction_longitudinal.py
+++ b/bench/contradiction_longitudinal.py
@@ -812,9 +812,9 @@ def write_outputs(
 
     env_with_duration = {**env, "wall_clock_duration_seconds": round(duration_seconds, 2)}
     json_blob = {"env": env_with_duration, "summary": summary}
-    json_path.write_text(json.dumps(json_blob, indent=2, default=str))
+    json_path.write_text(json.dumps(json_blob, indent=2, default=str), encoding="utf-8")
 
-    with csv_path.open("w", newline="") as fh:
+    with csv_path.open("w", newline="", encoding="utf-8") as fh:
         w = csv.writer(fh)
         w.writerow([
             "probe_id", "seed", "n_slice", "condition", "topic",
@@ -897,7 +897,7 @@ def write_outputs(
         "- **Metric B-contract (s4_contradiction hint OR anti_hits ≥80%)** tests what the system actually promises (REQUIREMENTS.md MEM-08, MCP-01 dual-route). Cosine cannot do either; pipeline either signals contradictions or it doesn't.",
         "",
     ]
-    md_path.write_text("\n".join(lines))
+    md_path.write_text("\n".join(lines), encoding="utf-8")
 
     print(f"[bench] Wrote: {md_path}")
     print(f"[bench] Wrote: {json_path}")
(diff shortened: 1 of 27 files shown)


CodeAbra:
Thanks for this, and for all the Windows digging you've done. Where it stands: v2.0 was a big rewrite, the whole storage and graph engine moved to Rust, so these diffs don't land clean on the new base anymore. But the bugs are real and still there on the new engine. Not letting them slip. We're porting your Windows fixes into the v2.0 daemon for a v2.0.3 patch. Keeping this open till it ships, then I'll close it with a link. Glad you ran it on Windows, this stuff is a pain to catch.

CodeAbra:
Ported into v2.0.3 (47f7175). pid-liveness now goes through psutil, and the crypto key file opens O_BINARY with an atomic replace. The auth-token piece doesn't have a v2.0 counterpart yet since the daemon transport is being reworked, so that part is tracked separately. Thanks for all the Windows digging, it landed.

PR #46, opened 2026-07-04
Fix lifecycle-lock acquire() race that allowed duplicate daemons

danielhertz1999-bit:
## Problem

`LifecycleLock.acquire()` used a **read → check → `os.replace()`** sequence with no atomicity (last-writer-wins). Two daemons launched in the same instant both read "no live lock", both wrote, and **both passed `acquire()` without raising `LifecycleLockConflict`**. The singleton guard in `daemon.main()` that is meant to make the loser exit (`return 1`) therefore never fired, so one daemon kept running as a **store-contending orphan**.

## Why it reproduces on Windows

Two independent launchers fire together:

1. The `iai-mcp-daemon` **scheduled task**, which runs the repo `.venv` interpreter (`<repo>\.venv\Scripts\pythonw.exe -m iai_mcp.daemon`).
2. The **wrapper/doctor respawn path** — `doctor._respawn_daemon()` spawns `[sys.executable, "-m", "iai_mcp.daemon"]`, and when triggered from a globally-installed `iai-mcp`, `sys.executable` resolves to a *different* (global) Python.

Because the Windows daemon binds an **ephemeral** port (`asyncio.start_server(..., "127.0.0.1", 0)` in `_ipc.start_ipc_server`), there is no bind collision to catch the duplicate either — both bind successfully, and `.locked` / `.daemon.port` / `.daemon.token` end up pointing at whichever process wrote last. Observed live: two `iai_mcp.daemon` processes started at the identical timestamp, one holding the published port, the other a portless orphan holding the store open for 13+ hours.

## Fix

Make the claim atomic:

- **`_write_exclusive()`** creates the lock with `os.open(..., O_CREAT | O_EXCL | O_WRONLY, 0o600)`. Exactly one racer can create the file; every other gets `FileExistsError` and concedes, so its daemon exits cleanly. Cross-platform (Windows + POSIX); preserves `0o600` (with an explicit `chmod` since `O_EXCL`'s mode is umask-subject on POSIX).
- **Safe stale-eviction**: a leftover lock (dead pid / corrupt / another host — the real dead-pid boot state) is evicted with an atomic `os.replace` that only one racer can win; a **post-steal liveness re-check** restores the file and concedes if a live daemon created it in the race window, so a real daemon's lock is never stolen. Bounded retry via `_ACQUIRE_MAX_ATTEMPTS`.

## Tests

- All existing lock tests preserved: stale recovery, corrupt-file overwrite, cross-host acquire, mode-0600, live-holder conflict.
- Adds **3 regression tests**, including a **12-thread concurrent-`acquire()` stress test** asserting exactly one winner and N-1 conflicts, from a seeded dead-pid stale lock (the real dual-launch boot state). All three **fail against the old `os.replace()` path** (the concurrent test even surfaces `PermissionError [WinError 5]` on Windows) and **pass with this change**.

No behavior change on the single-launcher path.

### DIFF
--- a/src/iai_mcp/lifecycle_lock.py
+++ b/src/iai_mcp/lifecycle_lock.py
@@ -23,6 +23,13 @@ DEFAULT_LOCK_PATH: Path = _default_lock_path()
 
 SCHEMA_VERSION: int = 1
 
+# acquire() serialises its read-check-write behind an OS advisory lock on a
+# sidecar guard file. How long to wait for that guard before giving up (a
+# healthy acquire() holds it for microseconds; a multi-second wait means a
+# peer is wedged mid-acquire).
+_GUARD_TIMEOUT_SEC: float = 10.0
+_GUARD_POLL_SEC: float = 0.02
+
 
 class LifecycleLockConflict(RuntimeError):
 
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks, this one's a nasty race and a good catch. Same situation as your other Windows PRs: v2.0 moved the engine to Rust so the diff doesn't apply to the new base as-is, but the lifecycle-lock race is still there on it, acquire is still read-check-replace instead of atomic. Porting the O_EXCL claim into v2.0 for the v2.0.3 patch. Keeping this open till it lands. Appreciate you chasing the dual-daemon thing down, that's hard to even reproduce.

CodeAbra:
Ported into v2.0.3 (47f7175). acquire() is an atomic O_CREAT|O_EXCL claim now, so the double-daemon window is closed on every platform, not just Windows. Good catch, thanks.

PR #47, opened 2026-07-05
fix(daemon): snapshot state before threaded save to end tick crash

Marsu6996:
### Problem

`save_state` is offloaded to a worker thread (`asyncio.to_thread(save_state, state)`) while the event loop keeps mutating the same `state` dict. `json.dump` iterating that live dict can raise:

```
tick failed: dictionary changed size during iteration
  daemon/__init__.py  json.dump(state, f, indent=2)
RuntimeError: dictionary changed size during iteration
```

The exception propagates out of the scheduler tick, so **consolidation silently stops**: episodes keep being captured, but they're never consolidated/indexed, and `memory_recall` only returns older, already-consolidated memories. It shows up under concurrent load (several MCP wrappers hitting the daemon at once), which is why it can look fine on a lightly-used setup and stall on a busy one.

### Fix

Add a small `_persist()` helper that takes an atomic snapshot in the event-loop thread before offloading:

```python
async def _persist(state: dict) -> None:
    snapshot = copy.deepcopy(state)          # synchronous in the loop -> no coroutine interleaves
    await asyncio.to_thread(save_state, snapshot)
```

`deepcopy` runs synchronously in the loop thread, so no other coroutine can mutate `state` mid-copy; the writer thread then serialises an isolated snapshot. All `save_state` offload sites route through it. State is JSON-serialisable so the copy is cheap.

### Test

`tests/test_persist_state_snapshot.py` — asserts the writer receives a deep copy (not the live dict) and that mutating `state` after the snapshot can't change what lands on disk. Fails if a call site reverts to passing the live `state`.

No version bump — leaving the release to you. Happy to rebase/trim or let you reimplement, whatever's least work.

---
🤖 Designed by Marsu — Refined by Claude

### DIFF
--- a/src/iai_mcp/daemon/__init__.py
+++ b/src/iai_mcp/daemon/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import copy
 import faulthandler
 import json
 import logging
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks. This was a real race, and the daemon got fully rewritten in v2.0 (the Rust move), so the tick-crash path here doesn't map directly, but snapshot-before-threaded-save is exactly the right idea. I'll confirm the new daemon holds that invariant and close this with a note if it's already covered. Keeping it open till I've checked. Solid catch on the dictionary-changed-size crash.

CodeAbra:
Superseded in v2.0. Every state artifact persists via mkstemp/O_EXCL, fsync, os.replace now, so the torn-file class is closed by idiom everywhere. Closing, but the invariant you pinned is exactly right. Thanks.

PR #48, opened 2026-07-05
Let daemon sleep with idle MCP wrappers

Marsu6996:
## Summary

This PR fixes a daemon sleep regression where idle MCP wrappers/control-plane probes could keep the daemon awake indefinitely, and adds an on-demand wake path for launchd-managed daemons after sleep/hibernation.

## Problem

Long-lived MCP sessions keep wrapper heartbeat files fresh, while watchdog/status probes arrive frequently. The daemon was treating those presence/control-plane signals as real activity, so the idle cycle could stay in WAKE and repeatedly do expensive wake-time work even when no memory operation was running.

There was also a recall hot path where the max-node community gate could scan the full assignment instead of only the current recall candidates.

## Changes

- Count only real memory activity for daemon idle decisions:
  - active/pending drains
  - recent JSON-RPC tool traffic
  - not wrapper heartbeat freshness or watchdog/status probes
- Wake the managed daemon from the MCP wrapper on `tools/call` when using the default macOS/Windows endpoint:
  - write `wake.signal` before kickstart
  - wait briefly for socket readiness
  - skip custom sockets and `IAI_MCP_DISABLE_DAEMON_AUTOSTART=1`
- Allow true WAKE transitions to bypass S2 anti-oscillation.
- Keep `RECALL_CONCURRENCY_DEFAULT=2` for multi-agent setups.
- Reduce recall CPU by scoring max-node community gates over current candidate embeddings only.
- Guard recall scoring/gating against NaN/Inf embeddings.
- Make daemon status prefer canonical `lifecycle_state.json` over stale legacy in-memory state, so `daemon status` does not report WAKE while the lifecycle FSM is DROWSY/SLEEP.

## Validation

Passed:

- `npm run typecheck && npm test` in `mcp-wrapper` — 51 passed
- `ruff check src/iai_mcp/concurrency.py src/iai_mcp/pipeline.py src/iai_mcp/socket_server.py tests/test_concurrency_session_open.py tests/test_mosaic_max_node_gate.py tests/test_recall_shared_cosine.py tests/test_socket_server_dispatch.py`
- targeted pytest recall/sleep/socket/cascade/status suite — 51 passed in the final de-duplicated pass; earlier broader targeted pass was 130 passed, 8 skipped before removing the #47 overlap from this branch

Live local smoke after installing the patch over the active daemon runtime:

- MCP wrapper initialized and listed 13 tools
- `memory_recall` returned non-error results
- canonical lifecycle state moved to DROWSY after the idle threshold
- daemon CPU dropped back to near-idle after the DROWSY edge work

Full local pytest was also run earlier. It reached 3640 passed, 122 skipped, 1 xfailed, with 2 local/environment failures unrelated to this diff:

- `tests/test_mosaicsigma_native_import.py::test_both_submodules_share_one_so` because the active site-packages directory contains several extension modules, not just `iai_mcp_native`.
- `tests/test_watchdog_phys_footprint.py::test_phys_footprint_not_above_rss_on_clean_process` under macOS memory-pressure/measurement conditions.

## Review

Reviewed with an aggressive Claude pass. Initial blockers were addressed:

- wake after hibernation is now guaranteed from the MCP wrapper hot path, not only at session start;
- the silent recall pool cap was removed;
- recall concurrency default remains 2;
- the status finishing pass is covered by targeted tests.

## Relationship with other open PRs

- #47 is our separate focused PR for the threaded `save_state` snapshot race. This PR no longer includes that fix.
- #44, #46, and #49 are separate Windows/lifecycle-lock/hippo-lock fixes and are not duplicates of this PR.

## Operational notes

- The first tool call after hibernation may wait up to 7 seconds for the daemon socket to return before falling through to existing fallback/error paths.
- The wrapper wake path is intentionally limited to the default managed user endpoint; custom socket deployments own their own supervision.

## Attribution

## CI follow-up

- macOS CI failed once on `tests/test_runtime_graph_centrality_child.py::test_parent_rss_lower_with_centrality_child` with `in_parent_delta=63.3MB` and `isolated_delta=65.0MB`.
- Local reruns showed the intended signal clearly (`168.5MB/49.4MB`, `167.1MB/47.9MB`), so the failure was a strict RSS allocator-noise assertion, not a runtime regression.
- The test now xfails only when RSS proof is inconclusive within a 32 MiB allocator-noise margin; it still fails if the child path is materially worse.

### DIFF
--- a/mcp-wrapper/package.json
+++ b/mcp-wrapper/package.json
@@ -13,7 +13,7 @@
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts",
     "typecheck": "tsc --noEmit",
-    "test": "node --import tsx --test test/*.test.ts"
+    "test": "node --import tsx --test --test-concurrency=1 test/*.test.ts"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.0.0",
--- a/mcp-wrapper/src/index.ts
+++ b/mcp-wrapper/src/index.ts
@@ -19,7 +19,10 @@ import {
   type ContentBlock,
   type SessionPayloadRaw,
 } from "./caching.js";
-import { WrapperLifecycle } from "./lifecycle.js";
+import {
+  shouldWakeDefaultDaemonForToolCall,
+  WrapperLifecycle,
+} from "./lifecycle.js";
 import {
   CONTEXT_EDITING_CONFIG,
   HOT_TOOLS,
@@ -40,6 +43,9 @@ export {
 };
 export type { ContentBlock, SessionPayloadRaw };
 
+interface ToolCallLifecycle {
+  ensureDaemonAlive(opts?: { waitForReachableMs?: number }): Promise<void>;
+}
 
 function toolResult(payload: unknown) {
   const content = [
@@ -57,6 +63,8 @@ function toolResult(payload: unknown) {
 export function buildServer(
   bridge?: PythonCoreBridge,
   spawnFn: typeof spawn = spawn,
+  lifecycle?: ToolCallLifecycle,
+  shouldWakeForToolCall: () => boolean = shouldWakeDefaultDaemonForToolCall,
 ): { server: Server; bridge: PythonCoreBridge } {
   const b = bridge ?? new PythonCoreBridge();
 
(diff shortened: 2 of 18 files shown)


CodeAbra:
Thanks. The idle-countdown logic got reworked in the v2.0 daemon so this diff doesn't apply as-is, but the 'presence isn't activity' point is the right one. Checking whether the new daemon already gates on real work only. Keeping it open till I confirm. Good thinking on the wrapper-heartbeat-vs-consolidation angle.

CodeAbra:
Superseded in v2.0. The idle source is last_activity_ts, stamped only for non-observation requests, so dashboard and observation polling can't reset the countdown. Your 'presence isn't activity' point is baked in. Closing, thanks.

PR #49, opened 2026-07-05
Windows: fix daemon wedge from blocking flock in hippo lock conversion

danielhertz1999-bit:
## Problem

On Windows the daemon periodically wedges: the liveness watchdog fires `pre_kill_forensic_dump` + kill, and the blackbox thread-dump shows the main thread stuck in `hippo/_db.py … downgrade_to_shared` → `_filelock.py … flock`. This has caused real capture outages (the daemon comes back only after a manual zombie-tree kill + restart).

## Root cause

`HippoDB.downgrade_to_shared` acquired the base lock with a **blocking** `flock(base_fd, LOCK_SH)` — the only lock site in `_db.py` without `LOCK_NB` (compare the shared-acquire and `escalate_to_exclusive` paths, both non-blocking + deadline-bounded).

On Windows, `_filelock.py` has no shared locks, so it services `LOCK_SH` as an **exclusive** `msvcrt` byte-range lock and has **no atomic EX→SH conversion** (its own docstring flags this hazard). A blocking `LOCK_SH` issued on an fd that *already* holds `LOCK_EX` makes the shim poll `LK_NBLCK` forever against the daemon's **own** range — Windows won't re-lock an already-locked region — so every thread stalls under the process-locks guard until the watchdog kills the process.

The daemon boots holding EXCLUSIVE and downgrades on its first WAKE tick, so this fires whenever it genuinely holds EXCLUSIVE at that point (intermittent: it no-ops early when the lock entry is already absent).

## Fix

- `downgrade_to_shared`: use `flock(base_fd, LOCK_SH | LOCK_NB)` and treat `EAGAIN`/`EWOULDBLOCK` as success. On Windows `SH == EX`, so the fd already owns the lock (relabel-only, no OS re-acquire); on POSIX an EX→SH downgrade never blocks anyway, so behavior is identical.
- Sibling self-conflict in `escalate_to_exclusive` (SH→EX on an already-held fd) — bounded by a 4s deadline so it *degraded* (raised `HippoLockHeldError`) rather than wedging, but still wrong on Windows. Fixed by exposing `SHARED_IS_EXCLUSIVE` from `_filelock.py` (True on Windows, False on POSIX) and short-circuiting to relabel-only when this fd already holds the lock. POSIX shared locks stay genuinely contended.

## Test

`tests/test_hippo_lock_cycle_no_wedge.py` runs the full daemon lock cycle (EXCLUSIVE → downgrade → escalate → downgrade), with every lock call — including `close()` — bounded by a watchdog-thread join timeout, so a reintroduced blocking flock **fails the assertion instead of hanging the suite**. It deliberately does **not** skip on Windows (unlike `test_hippo_concurrency.py`), which is exactly why this wedge went uncaught. Verified both ways: passes on the fix; on pre-fix code the downgrade test fails with the blocking-flock wedge.

Windows-only behavior change; POSIX paths are byte-for-byte equivalent.

### DIFF
--- a/src/iai_mcp/_filelock.py
+++ b/src/iai_mcp/_filelock.py
@@ -35,6 +35,13 @@ if platform.system() == "Windows":
     LOCK_NB = 4
     LOCK_UN = 8
 
+    # msvcrt has no shared locks, so LOCK_SH is serviced as an exclusive
+    # byte-range lock (see the module docstring). Callers doing in-place lock
+    # *conversion* need to know this: a "shared" hold here is already exclusive
+    # and cannot be co-held by another process, so escalating a shared lock this
+    # fd already owns is a relabel, not a contended acquire.
+    SHARED_IS_EXCLUSIVE = True
+
     _LOCK_BYTES = 2**30
     # Poll interval when emulating POSIX's block-until-acquired behaviour.
     _BLOCK_POLL_SECONDS = 0.05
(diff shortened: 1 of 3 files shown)


CodeAbra:
Thanks, this is a deep one. Same deal as the others, v2.0 is the Rust rewrite so the diff doesn't apply straight, but the wedge is real on the new engine, downgrade_to_shared still does a blocking flock that spins forever on Windows (no shared locks there). Porting the non-blocking LOCK_NB fix into v2.0 for v2.0.3. Keeping it open till then. Good detective work on the thread-dump, that's the kind of hang that's murder to trace.

CodeAbra:
Ported into v2.0.3 (47f7175). There's a compat seam now (_flock.py): POSIX re-exports fcntl, Windows maps onto an msvcrt region lock, so hippo imports and the lock conversion doesn't wedge. Thanks for chasing this down.

PR #51, opened 2026-07-05
Windows: watchdog self-kill must os._exit, not no-op sys.exit from a bg thread

danielhertz1999-bit:
## Problem

The liveness watchdog can detect a wedged daemon but fails to actually kill it on Windows, so the daemon stays down until a manual `taskkill`.

The watchdog runs in a background daemon thread (`name="iai-liveness-watchdog"`, `daemon/__init__.py`). When it decides to kill, `_self_kill()` in `daemon/_watchdog.py` branches on `hasattr(signal, "SIGKILL")`:

```python
if hasattr(signal, "SIGKILL"):
    os.kill(os.getpid(), signal.SIGKILL)   # POSIX: signals the whole process
else:
    sys.exit(1)                            # Windows: NO-OP from a bg thread
```

On Windows there is no `SIGKILL`, so it falls to `sys.exit(1)` — which only raises `SystemExit` in the **calling thread**. From the background watchdog thread that unwinds the thread, not the process, so the wedged daemon keeps running. Task Scheduler's `<MultipleInstancesPolicy>IgnoreNew</MultipleInstancesPolicy>` then treats the task as still "Running" and silently refuses to start a replacement, so the daemon stays down until someone kills the zombie by hand.

Observed in the wild: watchdog logs `daemon_wedge_kill pid=<pid>` (intent recorded) but the process with that pid is still alive minutes later.

## Fix

Use `os._exit(1)` on the no-`SIGKILL` path. It terminates the whole process immediately from any thread and bypasses interpreter shutdown — which is what you want here, since normal cleanup could itself hang on the same wedge the watchdog is escaping. POSIX behavior (`SIGKILL`) is unchanged.

## Test

The three existing `_self_kill` tests are all gated on `@_REQUIRES_SIGKILL`, so they **skip on Windows** — which is precisely why the no-op path shipped uncaught. The new test simulates the no-`SIGKILL` branch (via `monkeypatch.delattr(signal, "SIGKILL")`) so it runs on **every** platform, and asserts `_self_kill` calls `os._exit` and never `sys.exit`/`os.kill`. Verified both ways: passes on the fix, fails against the pre-fix source with `Failed: sys.exit is a no-op from the watchdog thread`.

Windows-only behavior change; POSIX paths are unaffected.

### DIFF
--- a/src/iai_mcp/daemon/_watchdog.py
+++ b/src/iai_mcp/daemon/_watchdog.py
@@ -593,7 +593,14 @@ def _self_kill(reason: str, kind: str) -> None:
     if hasattr(signal, "SIGKILL"):
         os.kill(os.getpid(), signal.SIGKILL)
     else:
-        sys.exit(1)
+        # Windows has no SIGKILL. The watchdog runs in a background thread
+        # (name="iai-liveness-watchdog", daemon=True), so sys.exit() would only
+        # raise SystemExit in *this* thread and leave the wedged daemon process
+        # alive -- and Task Scheduler's IgnoreNew policy then silently refuses to
+        # start a replacement. os._exit() terminates the whole process
+        # immediately from any thread, bypassing interpreter shutdown that could
+        # itself hang on the same wedge we are trying to escape.
+        os._exit(1)
 
 
 def _capture_blackbox(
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks, and good instinct on the bg-thread part. Same story, v2.0 is the Rust rewrite so the diff won't apply as-is, but the watchdog still does os.kill(getpid, SIGKILL) which is a no-op on Windows, so a wedged daemon just sits there. Porting the os._exit fix into v2.0 for v2.0.3. Keeping it open till it lands. Nice catch that sys.exit only unwinds the calling thread, that's an easy one to miss.

CodeAbra:
Ported into v2.0.3 (47f7175). The watchdog hard-exits via os._exit when SIGKILL isn't available. Clean fix, thanks.

PR #53, opened 2026-07-05
Windows: daemon stop must taskkill the whole process tree (/T)

danielhertz1999-bit:
## Problem

`iai_mcp daemon stop` on Windows can leave a child process alive and still holding the hippo DB lock, so a subsequent restart wedges.

The Windows stop path did:

```python
os.kill(pid, _signal.SIGINT)          # (1)
... poll until pid dies or timeout ...
if _is_pid_alive(pid):
    subprocess.run(["taskkill", "/F", "/PID", str(pid)])   # (2) no /T
```

Two problems compound:

1. On Windows `os.kill(pid, SIGINT)` is **not** a signal — Python maps any non-`CTRL_*` value to `TerminateProcess`, which hard-kills the **parent process only**. Any child the daemon spawned (a worker interpreter, a maturin build, etc.) is orphaned and keeps running — and keeps its handle on `~/.iai-mcp/hippo/.lock`.
2. Because `TerminateProcess` kills the parent immediately, the poll loop sees the pid gone and returns before the `taskkill` escalation runs. And even when it did run, `taskkill /F /PID` (no `/T`) only kills that one pid — by then childless.

Net effect: a "stopped" daemon can leave an orphan holding the lock, which then blocks the next daemon from acquiring it.

## Fix

Replace the SIGINT-then-escalate sequence with a single `taskkill /F /T /PID <pid>`. `/T` terminates the daemon **together with its child process tree**, while that tree is still intact. `schtasks /End` still runs first. This is a hard kill, but the old path was already a hard kill on Windows (`TerminateProcess`) — no graceful-shutdown behavior is lost, and the tree is now handled correctly.

POSIX (macOS/Linux) stop paths are untouched.

## Test

The Windows stop path previously had **zero** coverage (existing stop tests are macOS/Linux, or `@skipif` no-SIGKILL). The new test drives `cmd_daemon_stop` with `platform.system() == "Windows"` and asserts it issues `taskkill /F /T /PID` after `schtasks /End`, and never calls `os.kill` (which would orphan the tree). Verified both ways: passes on the fix, fails against the pre-fix source (which shows the `os.kill(SIGINT)` + `taskkill /F /PID` sequence).

### DIFF
--- a/src/iai_mcp/cli/_daemon.py
+++ b/src/iai_mcp/cli/_daemon.py
@@ -471,24 +471,18 @@ def cmd_daemon_stop(args: argparse.Namespace) -> int:
         payload = LifecycleLock().read()
         pid = payload["pid"] if payload else None
         if pid is not None and _is_pid_alive(pid):
-            try:
-                os.kill(pid, _signal.SIGINT)
-            except (ProcessLookupError, PermissionError) as exc:
-                logger.debug("SIGINT to daemon pid=%d failed: %s", pid, exc)
-                return 0
-
-            deadline = _time.monotonic() + _stop_escalation_bound()
-            interval = _stop_poll_interval()
-            while _time.monotonic() < deadline:
-                if not _is_pid_alive(pid):
-                    return 0
-                _time.sleep(interval)
-
-            if _is_pid_alive(pid):
-                _sp.run(
-                    ["taskkill", "/F", "/PID", str(pid)],
-                    check=False, capture_output=True,
-                )
+            # taskkill /T terminates the daemon together with any child processes
+            # it spawned (worker interpreters, maturin, etc.) as one tree, so none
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks. Same situation, v2.0 is the Rust rewrite so the diff won't apply straight, but the orphan problem is still live, stop still uses os.kill which on Windows only kills the parent and leaves a child holding the hippo lock, so the next start wedges. Porting the taskkill /F /T tree-kill into v2.0 for v2.0.3. Keeping it open till it lands. Good call spotting that the poll loop returned before the escalation even ran.

CodeAbra:
Ported into v2.0.3 (47f7175). Daemon stop has a real Windows branch now with taskkill /F /T so the store-lock child dies with the tree. Bonus, the getuid call that crashed Windows before any branch is fixed too. Thanks.

PR #54, opened 2026-07-05
Deflake centrality-child RSS isolation proof (best-of-N per arm)

danielhertz1999-bit:
## Problem

`test_parent_rss_lower_with_centrality_child` flakes on macOS CI. Recent failure:

```
AssertionError: centrality child isolation did not lower the parent footprint:
in_parent_delta=64.9MB isolated_delta=68.0MB
```

That's a ~3MB inversion on a ~65MB measurement — allocator noise, not a real regression. The test already flags itself as allocator-sensitive (`skipif(platform.system() != "Darwin", reason="settled-RSS isolation proof calibrated on this host's allocator")`), and it passes on the overwhelming majority of runs; a single noisy sample per arm is enough to flip the `isolated_delta < in_parent_delta` assertion.

## Fix

Settled-RSS noise is **one-sided**: transient allocations and allocator arena retention only ever *inflate* the measured resident delta — they never make an arm report *less* than it genuinely keeps resident. So the **minimum** delta across repeated identical builds is the cleanest estimate of each arm's true retained footprint.

Sample each arm `n_samples = 3` times (fresh store per rep, same deterministic corpus) and compare the minima. This removes the single-sample spikes that cause the intermittent inversion while preserving the proof's intent — the isolated arm must still genuinely retain less than the in-parent arm.

Pure test change; no production code is touched. The extra builds add a small amount of runtime to this one Darwin-only test.

## Verification

This test is `skipif != Darwin`, so it can't run off macOS — I verified the module compiles and still collects cleanly, and rely on this PR's macOS CI run for the behavioral check. If the isolation savings turn out to be genuinely marginal at `n_records=3000` (rather than noise), the min-of-3 comparison will surface that deterministically instead of flaking.

### DIFF
--- a/tests/test_runtime_graph_centrality_child.py
+++ b/tests/test_runtime_graph_centrality_child.py
@@ -516,15 +516,25 @@ def test_parent_rss_lower_with_centrality_child(store: MemoryStore):
     the detection child to also run in-process (so detection arenas AND the
     betweenness intermediate are reserved in the parent); the isolated arm uses
     the real spawn-context children for both. Each arm subtracts a fresh settled
-    baseline so it only measures the resident growth it itself retains.
+    baseline so it only measures the resident growth it itself retains, and is
+    sampled n_samples times with the minimum taken (see below) so one-sided
+    measurement noise can't invert the comparison.
     """
     import iai_mcp.community as _cm
 
     n_records = 3000
+    # Best-of-N de-noise: settled-RSS noise is one-sided — transient allocations
+    # and allocator arena retention only ever *inflate* the measured resident
+    # delta, never deflate it below what the arm genuinely keeps. So the minimum
+    # delta across repeated identical builds is the cleanest estimate of each
+    # arm's true retained footprint, and comparing the minima removes the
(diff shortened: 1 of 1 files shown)


CodeAbra:
Thanks. This one's a test-deflake, and the test itself changed a fair bit in v2.0 (the graph-cache and RSS-isolation paths got reworked in the Rust move), so the specific fix here is probably moot on the new base. Let me check whether the flake still shows up on the current test, if it does I'll carry the best-of-N idea over. Keeping it open till I've looked. Appreciate you tackling the flaky-RSS thing, allocator noise is the worst to pin down.

CodeAbra:
Closed as superseded. v2.0 reworked this test. Instead of an in-process settled-RSS delta (which flaked exactly as you describe, the allocator never returned the first arm's pages and inflated the second arm's baseline), each arm now runs in its own fresh process and reports peak RSS. That removes the one-sided noise architecturally, so the min-of-N sampling isn't needed on the new base. Your read on the flake was right, thanks.

PR #55, opened 2026-07-05
Stabilize daemon lifecycle after #47/#48 integration

Marsu6996:
## Draft / stacked PR

Do not merge before #47 and #48.

This PR is the daemon-lifecycle integration layer after #47/#48. It currently includes the #47 and #48 commits so the integration behavior can be validated together. After #47 and #48 merge, this branch will be rebased onto `main` so this PR only contains the net-new daemon stabilization commits.

Follow-up latency work has been split into draft stacked PRs:

- #57: recall latency (`ProcessType=Background` removal + cue embedding reuse)
- #58: topology latency (structural-result reuse now; further topology latency patch will land there before ready-for-review)

## Summary

Stabilize the daemon runtime behavior that surfaced once #47/#48 ran together:

- wrapper heartbeat presence no longer pins the daemon awake;
- stale sleep requests are normalized instead of replayed at boot;
- completed sleep cycles stay terminal unless real activity wakes the daemon;
- boot/sleep and active WAKE paths avoid heavy graph/session precache work;
- HIPPEA `session_open` reads warm structural cache instead of rebuilding the runtime graph;
- user-driven activity can wake immediately while non-user churn remains guarded;
- state persistence snapshots daemon state before threaded writes.

No public API change, no scheduler model rewrite, no release/tag.

## Review map

Suggested review order:

1. #47 snapshot save-state race
2. #48 idle wrapper wake/sleep behavior
3. this PR: integration stabilization on top of #47/#48
4. #57 recall latency
5. #58 topology latency

## Validation

Pre-split validation on this branch covered targeted daemon / recall / session / sleep / identity / socket suites, wrapper typecheck/tests, and live daemon checks. After splitting out the latency commit, GitHub CI should be treated as the authoritative check for this draft.

Local post-split pytest was not rerun because no compatible local Python environment currently has `pytest` installed.

## Attribution

### DIFF
--- a/mcp-wrapper/package.json
+++ b/mcp-wrapper/package.json
@@ -13,7 +13,7 @@
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts",
     "typecheck": "tsc --noEmit",
-    "test": "node --import tsx --test test/*.test.ts"
+    "test": "node --import tsx --test --test-concurrency=1 test/*.test.ts"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.0.0",
--- a/mcp-wrapper/src/index.ts
+++ b/mcp-wrapper/src/index.ts
@@ -19,7 +19,10 @@ import {
   type ContentBlock,
   type SessionPayloadRaw,
 } from "./caching.js";
-import { WrapperLifecycle } from "./lifecycle.js";
+import {
+  shouldWakeDefaultDaemonForToolCall,
+  WrapperLifecycle,
+} from "./lifecycle.js";
 import {
   CONTEXT_EDITING_CONFIG,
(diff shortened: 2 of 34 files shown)


CodeAbra:
Thanks, and this bundles #47/#48 which I'm already tracking. v2.0 rewrote the daemon lifecycle wholesale, so I'll fold the invariants you pinned here into the check against the new code. Keeping it open till I've confirmed what carried over. Appreciate the calmness-patch measurements.

CodeAbra:
Superseded in v2.0. The daemon lifecycle got rewritten wholesale and the invariants from #47/#48 hold by idiom. Closing this alongside them. Appreciate the measurements.

PR #56, opened 2026-07-05
Linux: Idle sensing feature similar to HID idleness in macOS

joerybka:
## Summary

Updates the idle sensor to read from the logind service on a Linux machine.  This aims to be agnostic between X11 and Wayland.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Build / tooling

## Affected areas

- [ ] Capture path
- [ ] Recall / retrieval
- [x] Daemon lifecycle / FSM
- [ ] Bench harness
- [ ] CLI / doctor
- [ ] Other: ___

## Testing

- [x] `pytest` passes locally
- [x] `ruff check src/ tests/` clean

## Benchmarks

This does not touch the active path of any benchmark-eligible change.

## Notes for reviewers

This change updates the idle sensor for Linux platforms to use logind, mimicking how the enriched check on macOS checks for HID idleness.  

### logind Differences
The underlying theory of the way that logind works on Linux and how IOHIDSystem work on macOS is different.  When querying IOReg on macOS, there is no need to scan for a session that is using a console, since there is only one console to worry about.  logind is predicated on multi-session, so the code must scan for all sessions using a console (seatN) and then look for the lowest idle time to return for actual comparison.  In truthfulness, this is me being extremely pedantic for completeness, the likelihood of this actually happening is very low.  

The current implementation could be pulled back and simplified to look for the first seat found (usually seat0), and then just use the IdleHint from that console.  Given the code already has to fall through the MCP heartbeat check to start polling for idleness, that would optimise it slightly, shaving about ~300 nanoseconds from the idle check.  I already have it written the way it is though, so not sure if you want me to pull that back or not.

I looked into unifying the idle approach, since they are both Unix-based, and that is actually a worse result with the same split complexity, because looking up the active terminals is different across platforms.

### DIFF
--- a/src/iai_mcp/doctor/_lifecycle_checks.py
+++ b/src/iai_mcp/doctor/_lifecycle_checks.py
@@ -524,17 +524,12 @@ def check_n_hid_idle_source() -> CheckResult:
     detector = IdleDetector()
     status = detector.status()
 
-    idle_str = (
-        f"{status.hid_idle_sec}s"
-        if status.hid_idle_sec is not None
-        else "unavailable"
-    )
     signals_str = (
         ",".join(status.available_signals) if status.available_signals else "none"
     )
 
     if "logind" in status.available_signals:
-        logind_str = (
+        idle_str = (
             f"{status.hid_idle_sec}s"
             if status.hid_idle_sec is not None
             else "not idle"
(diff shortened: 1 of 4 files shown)


joerybka:
@CodeAbra I have left this in draft, since I am looking for feedback on whether to simplify the idle check for the more common single-console case that this would likely be deployed with.

CodeAbra:
Thanks, and this one's a genuinely nice feature, not just a fix. It didn't make it into v2.0 (that release was the big Rust rewrite), but Linux idle sensing via logind is worth having, mirrors the macOS HID-idle check. I want to bring it into a v2.0.x minor. Keeping it open, I'll ping you when it's in. Really appreciate the thoroughness on the multi-seat handling, and no, definitely not holding 'real engineer' against you.

CodeAbra:
Merged into v2.2.0 (7de719d) with your authorship. The logind IdleHint/IdleSinceHint approach with per-seat aggregation (min across idle seats, not-idle if any seat is active) is exactly right, multi-seat pedantry included, and we kept it. Thanks for mirroring the macOS check so cleanly across X11 and Wayland.

PR #57, opened 2026-07-06
Reduce recall latency: macOS launchd throttling and cue embedding reuse

Marsu6996:
## Draft / stacked PR

Do not merge before #47, #48, and #55.

This PR is stacked after #55 (`local/runtime-pr47-pr48`). Until #55 lands, the GitHub diff may include dependency commits from the daemon branch. After #55 merges, this branch will be rebased/retargeted onto `main` so the review diff is only the recall-latency change.

Follow-up topology work is tracked separately in #58 and should not be merged before this PR.

## Summary

Reduce synchronous `memory_recall` latency in two small places:

- remove macOS launchd `ProcessType=Background` from both daemon plist sources, avoiding launchd background throttling for the socket daemon;
- reuse the cue embedding already computed by the ANN dispatch path instead of embedding the same cue again inside `recall_for_response`.

No API change, no scheduler rewrite, no topology changes.

## Why

Live diagnosis showed recall latency was already slow before any topology call. The two recall-side costs were launchd throttling and duplicate cue embedding on the hot path.

## Validation

- Not run locally after split: no compatible local Python environment currently has `pytest` installed.
- The pre-split combined branch had CI green before this commit was separated into stacked draft PRs.
- GitHub CI should be treated as the authoritative check for this draft.

## Attribution

### DIFF
--- a/mcp-wrapper/package.json
+++ b/mcp-wrapper/package.json
@@ -13,7 +13,7 @@
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts",
     "typecheck": "tsc --noEmit",
-    "test": "node --import tsx --test test/*.test.ts"
+    "test": "node --import tsx --test --test-concurrency=1 test/*.test.ts"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.0.0",
--- a/mcp-wrapper/src/index.ts
+++ b/mcp-wrapper/src/index.ts
@@ -19,7 +19,10 @@ import {
   type ContentBlock,
   type SessionPayloadRaw,
 } from "./caching.js";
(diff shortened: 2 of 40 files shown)


CodeAbra:
Thanks. The embedding-reuse idea is already in the v2.0 engine (the perf work went a similar direction), let me confirm your launchd-throttle piece is covered too. Keeping it open till I've checked. Recall latency is exactly the right thing to be chasing.

CodeAbra:
Confirmed in v2.0. Each dispatch embeds the cue once and the vector feeds every lane, so there are no repeated embeds to dedupe. Structural in the new engine. Closing as covered, thanks.

PR #58, opened 2026-07-06
Bound topology latency with cached snapshots

Marsu6996:
## Draft / stacked PR

Do not merge before #47, #48, #55, and #57.

This PR is stacked after the daemon and recall work. Until those dependencies land, the GitHub diff may include dependency commits from #55 and #57. After they merge, this branch will be rebased/retargeted onto `main` so the review diff is only the topology-latency work.

## Summary

Bound interactive topology latency without changing the background topology audit:

- reuse `assignment` and `rich_club` already returned by `build_runtime_graph()`;
- avoid recomputing C/L separately when `fast_sigma()` already returns sigma, C, and L;
- for large interactive topology requests, return the latest `sigma_observation` / `sigma_drift` event instead of rebuilding the runtime graph;
- return fast `insufficient_data` when a large store has no cached topology event yet;
- add topology provenance fields: `source`, `as_of`, and `age_s`;
- update the MCP topology schema so `sigma` is nullable and provenance fields are declared.

## Why

Live profiling showed interactive `topology` remained too slow on a large store because it rebuilt graph topology and recomputed shortest-path work synchronously. The hourly/background topology audit already writes the needed snapshot, so interactive callers should read that bounded result instead of repeating the heavy work.

## Behavior

- Small stores still compute live topology inline.
- Large stores return cached topology when available.
- Large stores without a cached snapshot return `source=insufficient`, `regime=insufficient_data` quickly.
- Background `compute_and_emit()` still performs the full topology computation and emits the snapshot events used by the interactive path.

## Validation

- `python3.11 -m py_compile` on changed Python modules/tests: pass
- `git diff --check`: pass
- `mcp-wrapper` `npm run typecheck`: pass
- `mcp-wrapper` `npm test`: 51 passed
- Python pytest not run locally after split because no compatible local Python environment currently has `pytest` installed; GitHub CI is the authoritative Python test run for this draft.

## Attribution

### DIFF
--- a/mcp-wrapper/package.json
+++ b/mcp-wrapper/package.json
@@ -13,7 +13,7 @@
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts",
     "typecheck": "tsc --noEmit",
-    "test": "node --import tsx --test test/*.test.ts"
+    "test": "node --import tsx --test --test-concurrency=1 test/*.test.ts"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.0.0",
--- a/mcp-wrapper/src/index.ts
+++ b/mcp-wrapper/src/index.ts
@@ -19,7 +19,10 @@ import {
   type ContentBlock,
   type SessionPayloadRaw,
 } from "./caching.js";
-import { WrapperLifecycle } from "./lifecycle.js";
+import {
+  shouldWakeDefaultDaemonForToolCall,
+  WrapperLifecycle,
+} from "./lifecycle.js";
 import {
   CONTEXT_EDITING_CONFIG,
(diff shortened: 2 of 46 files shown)


CodeAbra:
Thanks. Topology got reworked in v2.0, so I need to check whether the cached-snapshot bound you added is in the new path or worth porting. If it's not there it goes into a v2.0.x patch. Keeping it open till I confirm. Good instinct bounding that latency.

CodeAbra:
Superseded in v2.0. Topology serves from the memoized warm graph and persisted snapshot, O(1) on the warm path, nothing recomputes per call. Closing as covered, good instinct on the latency bound.

PR #60, opened 2026-07-07
Windows: consume_wake_signal drops wakes on transient sharing violation + de-flake race test

danielhertz1999-bit:
## Problem

Two issues surfaced by `tests/test_wake_handler.py::test_consume_atomic_no_race` failing on macOS CI with `assert 0 >= 1`.

**1. A real Windows bug in `consume_wake_signal()`.** It does an unlink-to-consume inside a bare handler:

```python
try:
    self._wake_signal_path.unlink()
except FileNotFoundError:
    return False
except OSError:          # <-- swallows PermissionError too
    return False
return True
```

On Windows, `unlink()` raises `PermissionError` (sharing violation, WinError 5/32) when a concurrent writer or reader momentarily holds the signal file open — Python's `open()` there doesn't request `FILE_SHARE_DELETE`. The bare `except OSError` swallows that transient failure as *"no signal,"* **dropping a real wake** and leaving the daemon stranded in HIBERNATION with an unserved socket. On POSIX this never happens (you can unlink an open file).

**2. The test itself is flaky.** `consumed_truthy_count >= 1` had no synchronization: the consumer ran a fixed 200 unlink iterations and could finish before any writer's first `replace` landed (adverse thread scheduling on a loaded runner), reading `consumed=0`. That's the `assert 0 >= 1` seen on macOS — independent of the Windows bug, and unfixed by a Windows-only change.

## Fix

- **`consume_wake_signal`**: retry the unlink briefly on a Windows `PermissionError` (mirrors `daemon_state._atomic_replace`). On POSIX a `PermissionError` is a genuine permission fault, so it still returns `False` on the first failure — no behavior change off Windows.
- **De-flake `test_consume_atomic_no_race`**: writers produce continuously and the consumer polls until it consumes (bounded by a deadline) instead of a fixed count. Keeps the atomicity intent (`errors == []`, no crash under concurrent write+consume) while making the `>=1` liveness check deterministic on any scheduler. The writer harness also retries `os.replace` on a Windows `PermissionError`, mirroring production.
- **Two deterministic regression tests** for the retry (simulated via monkeypatch, so they run on every platform): a transient Windows `PermissionError` is retried and consumed; a POSIX `PermissionError` is not retried.

## Verification

`wake_handler` suite green; `test_consume_atomic_no_race` looped 10× locally on Windows with zero flakes. The two new deterministic tests fail against the pre-fix source (the retry path doesn't exist), confirming they gate the fix. macOS behavior is exercised by this PR's CI.

### DIFF
--- a/src/iai_mcp/wake_handler.py
+++ b/src/iai_mcp/wake_handler.py
@@ -1,9 +1,24 @@
 from __future__ import annotations
 
+import platform
+import time
 from pathlib import Path
 
 
 __all__ = ["WakeHandler"]
+
+_IS_WINDOWS: bool = platform.system() == "Windows"
+
+# Windows os.unlink fails with PermissionError (WinError 5/ACCESS_DENIED or
+# 32/SHARING_VIOLATION) when another process momentarily holds the target
+# open -- e.g. the CLI/hook atomically replacing the signal, or `daemon
+# status`/the MCP server reading it, at the same instant the daemon consumes
+# it. Python's open() on Windows does not request FILE_SHARE_DELETE, so any
+# concurrent handle transiently blocks the unlink. The handle is held only
+# briefly, so a few short retries resolve it. POSIX unlink never sees this.
+# Mirrors iai_mcp.daemon_state._atomic_replace.
+_UNLINK_RETRY_ATTEMPTS: int = 10
+_UNLINK_RETRY_SLEEP_SEC: float = 0.05
 
 
 class WakeHandler:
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thanks. v2.0 (Rust rewrite) already has consume_wake_signal, so the wake-signal path carried over, I need to check whether the transient sharing-violation retry you added is in there or still missing. If it's missing it goes into v2.0.3 with the other Windows fixes. Keeping it open till I've confirmed. Nice de-flake on the race test too.

CodeAbra:
Checked in the shipped v2.0 code. consume_wake_signal catches the OSError (a Windows sharing-violation is a PermissionError subclass) and returns without consuming, so the signal survives and the next tick retries. Same invariant your fix added, held by cadence instead of an inline loop. Closing as covered, thanks for the careful report.

PR #62, opened 2026-07-08
daemon: auto-clear expired sleep quarantine at boot (re-surface of #36)

danielhertz1999-bit:
## Problem

Upstream auto-recovers an expired sleep quarantine only when a sleep cycle is *next attempted* (`_check_and_maybe_auto_recover_quarantine`, called at the top of the sleep-run path). A daemon that never reaches a sleep window — e.g. **pinned in WAKE with active sessions** — therefore carries a zombie quarantine entry indefinitely, and `doctor` check (l) keeps reporting a quarantine that expired days ago.

This is observable in the wild: a daemon has shown `doctor (l)` reporting a quarantine expired several days earlier while sitting in WAKE the whole time, because the recovery path was never exercised.

## Fix

Add `recover_expired_quarantine()` and call it once at daemon boot, so a long-expired quarantine is cleared eagerly instead of waiting for a sleep cycle that may not come. It:

- clears only an **expired** quarantine (an active one is left intact),
- is a no-op when there is no quarantine,
- is wrapped so boot recovery can never wedge startup.

This is purely additive (85 lines, no behavior change to the existing sleep-time auto-recovery) and self-contained.

## Tests

Three new tests in `tests/test_sleep_pipeline.py`:
- `test_recover_expired_quarantine_clears_without_sleep_cycle` — the core new behavior (clears without a sleep cycle running),
- `test_recover_expired_quarantine_keeps_active_quarantine` — an active quarantine is preserved,
- `test_recover_expired_quarantine_noop_when_absent` — no-op when none exists.

All quarantine tests (11) pass locally.

---

_Note: this fix was previously proposed as #36 (closed without landing). Re-surfacing it since the boot-recovery path is genuinely absent from `main` and the zombie-quarantine condition it fixes is live._


CodeAbra:
Thanks, and I remember #36. v2.0 has the quarantine handling, so I need to confirm whether the boot-time auto-clear made it into the new daemon or dropped in the rewrite. If it dropped it goes into v2.0.3. Keeping it open till I've checked. Good that you re-surfaced it, these boot-state things slip through easily.

CodeAbra:
Confirmed covered in v2.0. _check_and_maybe_auto_recover_quarantine clears an expired quarantine by TTL on the first consultation, including the first post-boot run, and handles malformed timestamps too. Closing as covered, and thanks for re-surfacing #36.

PR #65, opened 2026-07-11
Port v2.0.0 to Windows: Rust WAL/pager, Python IPC/filelock/crypto shims

danielhertz1999-bit:
Runs `v2.0.0` end-to-end on Windows 11. Verified live: daemon boots as v2.0.0 / WAKE via Task Scheduler, and the Node MCP wrapper round-trips `memory_recall` through the auth-gated TCP loopback path against a real Rust-engine store.

v2 dropped the v1.2.0 Windows work and the new Rust storage engine only compiled on Unix. This restores the port and extends it to the new engine. All changes are `#[cfg]` / `sys.platform` gated — POSIX behaviour is unchanged.

Happy to split into per-subsystem PRs if you'd prefer; kept it a single commit for reviewability of the whole story.

## Rust — `crates/lillibrain`

`wal.rs`, `journal.rs`, `pager.rs`, and one test used `std::os::unix::fs::FileExt::{read_exact_at, write_all_at}` unconditionally at 38 sites, so the crate did not compile on Windows at all.

- New `pos_io` module: cross-platform `PosIo` trait. Unix impl is a straight delegation to `std::os::unix::fs::FileExt`. Windows impl loops around `std::os::windows::fs::FileExt::{seek_read, seek_write}` to give the same `read_exact_at` / `write_all_at` semantics (Windows' primitives can short-transfer). Tests included.
- `wal.rs` / `journal.rs` / `pager.rs` / `tests/crash_recovery.rs`: swap the `use` for `crate::pos_io::PosIo`. Call sites are unchanged.

`cargo check --workspace --all-targets` is clean on Windows (warnings only).

## Python — shims and guards

Restored the two v1.2.0 shim modules verbatim from `2fc0296`:

- `src/iai_mcp/_filelock.py` — `fcntl.flock` on POSIX, `msvcrt.locking` with saved/restored offset on Windows.
- `src/iai_mcp/_ipc.py` — POSIX Unix-domain socket, Windows TCP loopback with a 32-byte random auth token file (icacls-restricted). Exposes `open_ipc_connection`, `start_ipc_server`, `make_sync_ipc_socket`, `send_sync_auth_token`, `shutdown_ipc`, etc.

Guarded top-level Unix-only imports so `import iai_mcp.*` doesn't crash on Windows:

- 6 files (`capture_queue`, `hippo/_db`, `hippo/__init__`, `lifecycle`, `lifecycle_event_log`, `lock_protocol`) swap `import fcntl` for `from iai_mcp import _filelock as fcntl`. Downstream `fcntl.flock(fd, fcntl.LOCK_EX)` calls work unchanged.
- `lillibrain/io.py`: gate `import fcntl` behind `_DARWIN` (only needed for `F_FULLFSYNC`).
- `daemon/__init__.py`: `try: import resource` → `None` on Windows, skip `_raise_fd_limit` there; filter shutdown-signal tuple with `getattr(signal, "SIGHUP", None)` so tuple construction doesn't `AttributeError` before the handler runs.

Routed AF_UNIX / `open_unix_connection` / `start_unix_server` through `_ipc` in 12 files (`brainview`, `cli/__init__`, `concurrency`, `core/__init__`, `daemon/_watchdog`, `direct_write`, `doctor/__init__`, `doctor/_lifecycle_checks`, `iai_cli`, `semantic_recall`, `socket_server`). Sync sites also call `send_sync_auth_token(sock)` immediately after `connect` so the Windows daemon's `_make_authenticated_handler` accepts the connection.

Other Windows-only quirks:

- `lifecycle_lock._is_pid_alive`: skip `os.kill(pid, 0)` on Windows (it raises `OSError [WinError 87]` — invalid parameter — for a live process) and fall through to the existing `psutil.pid_exists` path. The 5 other `os.kill(pid, 0)` sites (capture, `cli/_maintenance`, `doctor/_lifecycle_checks`, `heartbeat_scanner`, `hippo/__init__`) are not on the boot path — happy to add if you want them in the same PR.
- `crypto._try_file_get`: gate the Unix `st_mode & 0o077` mode check and the `st_uid != os.geteuid()` check behind `sys.platform != "win32"` (Windows enforces access via ACLs, not mode bits).
- `crypto._try_file_set`: add `os.O_BINARY` to the open flags (Windows text-mode translates raw `0x0A` bytes in the AES key to `0x0D 0x0A`, silently corrupting it), `getattr` around `os.fchmod` (does not exist on Windows), and use `os.replace` instead of `os.rename` (Windows `rename` refuses to overwrite).

## Node MCP wrapper

- Restore `mcp-wrapper/src/ipc.ts` with `getDaemonConnectTarget`, `daemonUnreachableHint`, `createDaemonConnection`, `readDaemonToken`, and `authenticateConnection`. Includes the `IAI_DAEMON_TCP_PORT` / `IAI_DAEMON_TCP_HOST` test escape hatch so the TCP+auth path can be exercised without depending on `process.platform`.
- `bridge.ts`: route all three connect sites (`_doStart`, reconnect handler, `emitSessionOpen`) through the new helpers and call `authenticateConnection(sock, target)` immediately after connect. Without this the Windows daemon's `_make_authenticated_handler` silently closes every connection.

## Verification

- `cargo check --workspace --all-targets` on Windows 11 — clean (warnings only).
- `pip install -e .` in a Python 3.12 venv — native extension builds and imports.
- `iai-mcp daemon status` reports `version: 2.0.0`, `state: WAKE`, TCP port `54597`, uptime healthy.
- `iai-mcp bank-recall --query "iai" --limit 2` returns real records from a pre-existing store that auto-migrated from v1 hnswlib to the v2 Rust engine on first boot.
- Node stdio round-trip: `initialize` + `notifications/initialized` + `tools/call memory_recall` returns hits from the store with `reason: "exact-cosine"` (Rust cosine path).

## Not included

- Test-suite port for the 380+ new v2 tests. v1.2.0 shipped a Windows-friendly test suite as part of `2fc0296`; that scope hasn't been redone yet against v2's much-larger test tree. Runtime is proven, but CI would still need work.
- Task Scheduler daemon-install-path refactor in `cli/_daemon.py`. Existing scheduled task from v1.x runs the same `python -m iai_mcp.daemon` entry, which works on v2, so this didn't block me — but a fresh Windows install would need the installer path restored.
- The 5 non-boot `os.kill(pid, 0)` sites noted above.

If any of these should be in this PR let me know and I'll extend it. Also happy to split by subsystem — kept one commit here so the story reads front-to-back, but you may prefer separate PRs for Rust vs Python vs Node.

Co-authored-by: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/crates/lillibrain/src/journal.rs
+++ b/crates/lillibrain/src/journal.rs
@@ -16,11 +16,11 @@
 //! ```
 
 use std::fs::{File, OpenOptions};
-use std::os::unix::fs::FileExt;
 use std::path::{Path, PathBuf};
 
 use crate::consts::PAGE_SIZE;
 use crate::error::Result;
+use crate::pos_io::PosIo;
 
 /// Journal magic: exactly 12 bytes.
 pub const JOURNAL_MAGIC: [u8; 12] = *b"LILLI-JRNL\x00\x00";
--- a/crates/lillibrain/src/lib.rs
+++ b/crates/lillibrain/src/lib.rs
@@ -18,6 +18,7 @@ pub mod gates;
 pub mod io;
 pub mod journal;
 pub mod pager;
+pub mod pos_io;
 pub mod py;
 pub mod store;
 pub mod wal;
--- a/crates/lillibrain/src/pager.rs
+++ b/crates/lillibrain/src/pager.rs
@@ -14,11 +14,12 @@
 
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
-use std::os::unix::fs::FileExt;
 use std::path::{Path, PathBuf};
 
 use fs4::FileExt as Fs4FileExt;
 use parking_lot::Mutex;
+
+use crate::pos_io::PosIo;
 
 use crate::consts::{
     DB_MAGIC, HDR_DB_SIZE_OFFSET, HDR_FIRST_FREELIST_OFFSET, HDR_FREELIST_COUNT_OFFSET,
--- a/crates/lillibrain/src/pos_io.rs
+++ b/crates/lillibrain/src/pos_io.rs
@@ -1,0 +1,118 @@
+//! Cross-platform positional file I/O.
+//!
+//! The WAL, journal, and pager code writes and reads fixed-size pages at known
+//! offsets. On Unix that's spelled `FileExt::{read_exact_at, write_all_at}`.
+//! Windows has `seek_read` / `seek_write`, but they can short-read/short-write
+//! and they don't provide read_exact / write_all semantics on their own.
+//!
+//! `PosIo` gives every platform the read_exact_at / write_all_at surface the
+//! rest of the crate expects. The Unix impl is a straight delegation. The
+//! Windows impl loops around `seek_read` / `seek_write`, matching the Unix
+//! trait's short-transfer behaviour so callers can treat both platforms the
+//! same.
+
+use std::fs::File;
+use std::io::{Error, ErrorKind, Result};
+
+pub trait PosIo {
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> Result<()>;
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> Result<()>;
+}
+
+#[cfg(unix)]
+impl PosIo for File {
+    #[inline]
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> Result<()> {
+        std::os::unix::fs::FileExt::read_exact_at(self, buf, offset)
+    }
+
+    #[inline]
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> Result<()> {
+        std::os::unix::fs::FileExt::write_all_at(self, buf, offset)
+    }
+}
+
+#[cfg(windows)]
+impl PosIo for File {
+    fn read_exact_at(&self, mut buf: &mut [u8], mut offset: u64) -> Result<()> {
+        while !buf.is_empty() {
+            match std::os::windows::fs::FileExt::seek_read(self, buf, offset) {
+                Ok(0) => break,
+                Ok(n) => {
+                    let tmp = buf;
+                    buf = &mut tmp[n..];
+                    offset += n as u64;
+                }
+                Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+                Err(e) => return Err(e),
+            }
+        }
+        if buf.is_empty() {
+            Ok(())
+        } else {
+            Err(Error::new(
+                ErrorKind::UnexpectedEof,
+                "failed to fill whole buffer",
+            ))
+        }
+    }
+
+    fn write_all_at(&self, mut buf: &[u8], mut offset: u64) -> Result<()> {
+        while !buf.is_empty() {
+            match std::os::windows::fs::FileExt::seek_write(self, buf, offset) {
+                Ok(0) => {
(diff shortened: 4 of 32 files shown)


danielhertz1999-bit:
Two follow-up commits added (fast-forward, no rebase):

- `ba876bb` — the same `os.kill(pid, 0)` → `sys.platform != "win32"` guard applied to `doctor/_lifecycle_checks.py::check_a_daemon_alive` and `heartbeat_scanner.py::_is_pid_alive` (both hit WinError 87 on every doctor run before this). Also swaps the residual `import fcntl as _fcntl` inside `doctor.check_c_lock_healthy` for the `_filelock` shim.
- `c605a1a` — fixes the 4 macOS CI regressions the initial commit caused. `brainview.py` and `iai_cli.py` connect to `str(self._socket_path())` / `str(_daemon_socket_path())`, which are the exact hook points `test_brainview` and `test_cli_daemon_relay` monkeypatch to point at a fake daemon. The initial port swapped both for `make_sync_ipc_socket()` and lost the injected path. This commit adds an optional `addr=` override on `make_sync_ipc_socket` (ignored on Windows, where the port file is authoritative) and threads the original per-caller paths through it. POSIX bytes on the wire are now identical to pre-port.

## CI triage — the 17 failures that survive are pre-existing v2.0.0 issues, not from this PR

Missing dev-deps in the CI image:
- `test_bench_lme_blind_preflight.py` × 5 — `RuntimeError: huggingface_hub not installed`
- `test_daemon_fdlimit_and_fsm.py::test_rendered_plist_contains_fd_floor` — `ModuleNotFoundError: No module named 'defusedxml'`

Thread-leak / lifecycle regressions:
- `test_hippo_storage::test_close_drains_async_writes_without_explicit_disable` — `iai-mcp-async-writes` daemon thread leaks past `close()`
- `test_reinforce_defer::test_no_thread_leak_after_disable` — 8 `_idle_janitor` threads + `iai-mcp-reinforce-queue` outlive disable

Perf / determinism / data-model:
- `test_lilliengine_gil_concurrency::test_engine_scan_releases_gil` — 48.9 ms GIL gap vs 30 ms budget
- `test_exact_authority_index::test_build_top_k_matches_brute_force_reference` — ANN id mismatch
- `test_runtime_graph_cache_pending_key` × 2 — cache-key tuple shape (`5` vs `6` entries)
- `test_semantic_degraded_path::test_pending_row_excluded_from_warm_semantic_until_reembed`

Runner-specific / brittle:
- `test_mosaic_gpl_removal::test_leiden09_perf_post_removal` — expects `21-09-SUMMARY.md` written by a separate task
- `test_spatial_scaffold` × 3 — asserts path root is `'IAI-MCP'` but the runner sees `'src'`

None of these touch anything modified in this PR (Rust WAL/pager, `_ipc`, `_filelock`, or the guards). Happy to open separate issues for whichever are useful; not fixing them here to keep scope tight to the port itself.

CodeAbra:
Merged into v2.1.0 (6bfc5d9), with your authorship on the port commit. This is the whole engine on Windows: your pos_io seam got lillibrain compiling there, and your TCP-loopback + token transport closed the AF_UNIX gap the async daemon had on Windows. We reconciled the few files that overlapped our landed v2.0.3 Windows patch in favor of those versions, and hardened the token handshake (ACL before the secret is written). Task Scheduler install and the 380+ Windows test port are tracked for a follow-up (v2.1.1). Genuinely impressive work, thank you.

PR #73, opened 2026-07-13
Complete the Windows port on top of the v2.0.3 baseline

danielhertz1999-bit:
### Problem

v2.0.3 fixed the Windows daemon lifecycle path (`_flock` msvcrt shim, atomic
lifecycle-lock claim, crypto key WRITE path). Four gaps remained that keep
`pip install -e .` and daemon runtime from working end-to-end on Windows:

- `crates/lillibrain` calls `std::os::unix::fs::FileExt` at 38 sites (WAL,
  journal, pager, `tests/crash_recovery`). Rust build fails to compile on
  Windows.
- The Node MCP wrapper (`mcp-wrapper/src/bridge.ts`) calls
  `open_unix_connection` directly at three sites; no auth-token handshake and
  no way to reach the daemon over TCP loopback.
- Python throughout the package calls `asyncio.open_unix_connection` /
  `asyncio.start_unix_server`. Windows has no AF_UNIX; these fail at every
  boundary.
- `crypto.py` still crashes on the READ path — `os.stat().st_uid !=
  os.geteuid()` calls `os.geteuid` which does not exist on Windows.
- `doctor/_lifecycle_checks.py check_a_daemon_alive` still calls
  `os.kill(pid, 0)`, which returns OSError [WinError 87] against a live pid
  and makes every doctor run report the daemon dead.

### Solution

Five commits, additive on top of `main`:

1. **Rust cross-platform `PosIo` trait** in a new `crates/lillibrain/src/pos_io.rs`.
   POSIX delegates to `FileExt`; Windows loops `seek_read` / `seek_write` to
   preserve `read_exact` / `write_all` semantics. `wal.rs`, `journal.rs`,
   `pager.rs`, and `tests/crash_recovery.rs` switch from `FileExt` to `PosIo`.

2. **Node MCP wrapper TCP + auth-token handshake.** New
   `mcp-wrapper/src/ipc.ts` provides `getDaemonConnectTarget` (prefers
   `~/.iai-mcp/.daemon.{port,token}` when present, falls back to the AF_UNIX
   path) plus `authenticateConnection`. `bridge.ts`'s three connect sites are
   rewired.

3. **Python AF_UNIX → TCP loopback fallback** in a new `src/iai_mcp/_ipc.py`,
   with an authentication handshake mirroring the Node wrapper's. Sync-mode
   helpers preserve test-injected daemon socket paths. Fifteen Python modules
   swap `asyncio.open_unix_connection` / `start_unix_server` for
   `_ipc.open_ipc_connection` / `_ipc.start_ipc_server`: `brainview`,
   `capture_queue`, `cli/__init__`, `concurrency`, `core`, `daemon/__init__`,
   `direct_write`, `doctor/__init__`, `heartbeat_scanner`, `iai_cli`,
   `lifecycle`, `lifecycle_event_log`, `lillibrain/io`, `semantic_recall`,
   `socket_server`.

4. **`crypto.py` READ-path Windows guards.** `stat.st_mode & 0o077` and the
   `os.geteuid()` comparison are gated on `sys.platform != 'win32'`. Access
   control on Windows stays enforced via ACLs at write time (v2.0.3's O_BINARY
   + os.replace + hasattr-guarded fchmod).

5. **`doctor/_lifecycle_checks.py` liveness and lock-shim swaps.**
   `check_a_daemon_alive` gates the raw `os.kill(pid, 0)` on
   `sys.platform != 'win32'`; the psutil verify below is the source of truth
   on Windows. `check_c_lock_healthy` swaps its local `import fcntl` for
   `from iai_mcp import _flock as _fcntl`. `daemon/_watchdog.py`
   `_probe_status_roundtrip` routes through `_ipc.open_ipc_connection`.

### Guarantees and scope

POSIX behavior is unchanged. The Rust `PosIo` trait delegates to `FileExt`
verbatim on Unix. The Python `_ipc` shim falls back to the existing
`asyncio.open_unix_connection` when the TCP token file is absent (i.e.
POSIX daemons keep their existing AF_UNIX transport). `_flock.py` (from
v2.0.3) already re-exports `fcntl.flock` verbatim on POSIX. Crypto and
doctor guards are guarded by explicit `sys.platform` checks.

No change to protocol, schema, or on-disk formats.

### Validation

On Windows 11 (Python 3.12.10, Rust 1.87, VS Build Tools):

- `cargo check --workspace` clean (Rust extension builds).
- All eight Windows-touched Python modules import cleanly
  (`_flock`, `_ipc`, `crypto`, `lifecycle_lock`, `hippo`, `hippo._db`,
  `lock_protocol`, `doctor._lifecycle_checks`, `daemon._watchdog`).
- `iai-mcp doctor`: 24/26 PASS. Two pre-existing FAILs on Windows unrelated
  to this PR: (b) socket-file check probes the AF_UNIX path even in TCP mode,
  and (q) reports the `iai` alias not on PATH.
- Live `iai-mcp bank-recall --query "windows port"` returns records via the
  running daemon end-to-end (Node wrapper → TCP → daemon → SQLite → HNSW).
- `iai_mcp.daemon` boots via existing Task Scheduler entry as v2.0.0/WAKE and
  runs a full sleep/wake cycle.

Not run in this PR: the full pytest suite on Windows (this port targets
runtime; test-suite port is a separate follow-up).

### Known open items (deferred)

- Remaining raw `os.kill(pid, 0)` sites in `capture.py`, `cli/_maintenance.py`,
  `hippo/__init__.py`. Correctness held by the flock; worth porting to the
  v2.0.3 `lifecycle_lock._is_pid_alive` psutil-first pattern for consistency.
- `doctor` check_b — currently probes AF_UNIX socket file; would need a
  platform branch or a "reachable-via-config" probe that reads `.daemon.port`.
- Task Scheduler install-path refactor in `cli/_daemon.py`.
- Windows-specific test-suite port (v2 has 900+ new tests).

### DIFF
--- a/crates/lillibrain/src/journal.rs
+++ b/crates/lillibrain/src/journal.rs
@@ -16,11 +16,11 @@
 //! ```
 
 use std::fs::{File, OpenOptions};
-use std::os::unix::fs::FileExt;
 use std::path::{Path, PathBuf};
 
 use crate::consts::PAGE_SIZE;
 use crate::error::Result;
+use crate::pos_io::PosIo;
 
 /// Journal magic: exactly 12 bytes.
 pub const JOURNAL_MAGIC: [u8; 12] = *b"LILLI-JRNL\x00\x00";
--- a/crates/lillibrain/src/lib.rs
+++ b/crates/lillibrain/src/lib.rs
@@ -18,6 +18,7 @@ pub mod gates;
 pub mod io;
 pub mod journal;
 pub mod pager;
+pub mod pos_io;
 pub mod py;
 pub mod store;
 pub mod wal;
--- a/crates/lillibrain/src/pager.rs
+++ b/crates/lillibrain/src/pager.rs
@@ -14,11 +14,12 @@
 
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
-use std::os::unix::fs::FileExt;
 use std::path::{Path, PathBuf};
 
 use fs4::FileExt as Fs4FileExt;
 use parking_lot::Mutex;
+
+use crate::pos_io::PosIo;
 
 use crate::consts::{
     DB_MAGIC, HDR_DB_SIZE_OFFSET, HDR_FIRST_FREELIST_OFFSET, HDR_FREELIST_COUNT_OFFSET,
--- a/crates/lillibrain/src/pos_io.rs
+++ b/crates/lillibrain/src/pos_io.rs
@@ -1,0 +1,118 @@
+//! Cross-platform positional file I/O.
+//!
+//! The WAL, journal, and pager code writes and reads fixed-size pages at known
+//! offsets. On Unix that's spelled `FileExt::{read_exact_at, write_all_at}`.
+//! Windows has `seek_read` / `seek_write`, but they can short-read/short-write
+//! and they don't provide read_exact / write_all semantics on their own.
+//!
+//! `PosIo` gives every platform the read_exact_at / write_all_at surface the
+//! rest of the crate expects. The Unix impl is a straight delegation. The
+//! Windows impl loops around `seek_read` / `seek_write`, matching the Unix
+//! trait's short-transfer behaviour so callers can treat both platforms the
+//! same.
+
+use std::fs::File;
+use std::io::{Error, ErrorKind, Result};
+
+pub trait PosIo {
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> Result<()>;
+    fn write_all_at(&self, buf: &[u8], offset: u64) -> Result<()>;
+}
+
+#[cfg(unix)]
+impl PosIo for File {
+    #[inline]
+    fn read_exact_at(&self, buf: &mut [u8], offset: u64) -> Result<()> {
+        std::os::unix::fs::FileExt::read_exact_at(self, buf, offset)
+    }
+
+    #[inline]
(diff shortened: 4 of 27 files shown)


danielhertz1999-bit:
Heads-up on the CONFLICTING status: this branch was reconciled and verified on top of v2.0.3, but upstream has since released v2.1.0 → v2.2.2 (30 commits, including further Rust WAL/journal/pager work in `crates/lillibrain` that clashes with the PosIo commit here).

Happy to rebase onto current `main` once someone from upstream signals it's worth carrying forward — but I'd rather get feedback on the overall shape of the port first (Rust `PosIo` trait, Node `_ipc.ts` + auth handshake, Python `_ipc.py` AF_UNIX→TCP fallback, `sys.platform`-gated crypto/doctor guards) before spending time on a v2.2.2-baseline rebase that may need the shape adjusted anyway.

All commits are independent and can be picked individually if part of the port is more useful than the whole.

danielhertz1999-bit:
Closing — just noticed commit 6bfc5d9 (Add Windows support: Rust positional-I/O seam and TCP-loopback daemon transport, merged in v2.1.0) is essentially this same port under my authorship, plus commit 9c73027 (Route every liveness probe through the canonical pid_exists) closes the deferred `os.kill(pid, 0)` items I listed. This PR is redundant against v2.2.2.

Thanks to Areg for the follow-up polish. No action needed on your side.

PR #74, opened 2026-07-14
Windows: restore daemon boot shims + Task Scheduler lifecycle (v2.3.0 still can't start on Windows)

danielhertz1999-bit:
## Summary

On Windows, the daemon **cannot start at all** on current `main` (v2.3.0): `python -m iai_mcp.daemon` dies during import/boot on a chain of POSIX-only calls, and `iai-mcp daemon install`/`start` have no Windows path. This PR restores the cross-platform shims (which existed in the v1 Windows port but were dropped in the v2.0.0 rewrite and never re-added) and guards the remaining POSIX-only sites, so the daemon boots and is manageable on Windows again.

All changes are **no-ops on macOS/Linux** — they are guarded by `sys.platform`/`os.name`/`hasattr`/`getattr` checks or route through a shim that delegates to the original stdlib call on POSIX.

## What was broken on Windows

Starting the daemon crashed in sequence on:

1. `daemon/__init__.py` — top-level `import resource` (POSIX-only → `ModuleNotFoundError`)
2. `daemon/__init__.py` — `signal.SIGHUP` referenced in an eagerly-evaluated tuple (`AttributeError` before the surrounding `try/except` runs)
3. `socket_server.py` — `inspect.signature(asyncio.start_unix_server)` evaluated *before* the existing `IS_WINDOWS` branch that already avoids it (`AttributeError`)
4. `capture_queue.py`, `lifecycle.py`, `lifecycle_event_log.py`, `lillibrain/io.py` — unconditional `import fcntl` (the `_filelock.py` shim they used was deleted in the v2.0.0 rewrite `1a333da`)
5. `doctor/_lifecycle_checks.py::check_c_lock_healthy` — local `import fcntl`
6. `cli/_daemon.py` — `daemon install`/`uninstall`/`start` call `os.getuid()` before any platform branch and have no Windows branch (`install`/`uninstall` fall into Linux-only `systemctl` logic; `start` prints "Unsupported OS")
7. `cli/_crypto.py::cmd_crypto_status` — unguarded `os.geteuid()` (`AttributeError`)
8. `capture.py`, `migrate/_to_lilli_verify.py` — `os.fchmod` not guarded (`AttributeError`; sibling sites in `crypto.py`/`memory_bank.py` already guard it)

## Changes

**Restore the file-locking shim (commit 1)**
- Re-add `src/iai_mcp/_filelock.py` — a cross-platform `flock`-compatible shim (`LockFileEx`/`UnlockFileEx` on Windows, delegates to stdlib `fcntl` on POSIX), imported as `from iai_mcp import _filelock as fcntl` so call sites are unchanged.
- Repoint `capture_queue.py`, `lifecycle.py`, `lifecycle_event_log.py`, `lillibrain/io.py`, `doctor/_lifecycle_checks.py` at the shim.
- `daemon/__init__.py`: make `import resource` local + gated on `sys.platform == "win32"` inside `_raise_fd_limit()` (fd-limit raising is a POSIX-only concept — early-returns on Windows); guard `signal.SIGHUP` with `getattr(signal, "SIGHUP", None)`.
- `socket_server.py`: move the `inspect.signature(asyncio.start_unix_server)` call below the existing `IS_WINDOWS` branch.

**Task Scheduler daemon management + remaining guards (commit 2)**
- `cli/__init__.py`: add `_is_windows()` and `SCHTASKS_TASK_NAME`.
- `cli/_daemon.py`: add `_find_pythonw()` + `_render_schtasks_xml()`, and real Windows branches to `cmd_daemon_install`/`uninstall`/`start` (`schtasks /Create`+`/Run`, `/End`+`/Delete`, `/Run`). Guard the macOS/Linux `os.getuid()` calls with `hasattr(os, "getuid")`. (`daemon stop` already had its `taskkill /F /T` branch.)
- `cli/_crypto.py`: gate `os.geteuid()` behind `os.name != "nt"` (mirrors the read-path guard in `crypto.py`); report `mode_secure`/`uid_matches_process` as `null` on Windows, where `st_mode`/`st_uid` are synthetic and access is governed by ACLs.
- `capture.py`, `migrate/_to_lilli_verify.py`: guard `os.fchmod` with `hasattr(os, "fchmod")`, matching the existing `crypto.py`/`memory_bank.py` sites.

## Verification (Windows 11, Python 3.12)

- `python -m iai_mcp.daemon` boots to `WAKE` and reports `version: 2.2.2` (verified on the v2.2.2 base these commits were authored against; rebased cleanly onto v2.3.0 here — all 13 files byte-compile on the new base).
- `iai-mcp daemon status` → healthy; `iai-mcp doctor` passes except the pre-existing `.daemon.sock`-absent check (Windows uses the TCP+auth-token transport, not a Unix socket).
- `iai-mcp crypto status`, `iai-mcp daemon install --dry-run`, `iai-mcp daemon start` all run cleanly (each previously crashed).
- Targeted CLI test suite (`test_cli_daemon.py` + `test_cli_crypto.py`) goes **26 → 15 failures** on a Windows host (fixed 11, zero regressions vs. the pre-patch baseline). The 15 remaining are pre-existing POSIX-specific *test assertions* running on Windows (e.g. asserting mode `0o600` where Windows reports `0o666`, mocked `launchctl` paths) — a separate test-suite-port concern, not addressed here.
- A whole-tree sweep of six POSIX-only bug classes (uid family, `fchmod`/`chown`, `fcntl`/`resource` imports, POSIX signals + `os.kill`, `AF_UNIX` sockets, `fork`/`setsid`/`rlimit`) found no further unguarded sites after these changes.

## Notes

- No behavior change on macOS/Linux — every change is platform-guarded or shim-delegated.
- Happy to split into two PRs (boot shims / daemon-lifecycle) if you'd prefer to review them separately.

### DIFF
--- a/src/iai_mcp/_filelock.py
+++ b/src/iai_mcp/_filelock.py
@@ -1,0 +1,100 @@
+"""Platform-agnostic file locking shim.
+
+On POSIX: thin wrapper around fcntl.flock.
+On Windows: msvcrt.locking with errno normalisation so callers checking
+errno.EWOULDBLOCK / errno.EAGAIN on non-blocking failures work unchanged.
+The file offset is saved/restored around each call (msvcrt locks relative to
+the file position; fcntl.flock does not move it) and the blocking path polls
+so it waits indefinitely like POSIX rather than giving up after msvcrt's ~10 s.
+
+Known divergence — shared locks are not truly shared on Windows.
+``msvcrt.locking`` only offers exclusive byte-range locks, so LOCK_SH is
+serviced as an exclusive lock: a second concurrent reader blocks where POSIX
+would let both in. This is a throughput limitation, not a correctness one, and
+it is deliberately NOT fixed with Win32 ``LockFileEx`` (which does support
+shared locks) because callers in ``hippo/_db.py`` rely on fcntl.flock's atomic
+lock *conversion* — downgrading EXCLUSIVE->SHARED and escalating SHARED->
+EXCLUSIVE in place on the same fd. ``LockFileEx`` has no atomic conversion
+(you must Unlock then re-Lock, racing other waiters), so swapping it in would
+trade a throughput limit for a correctness hazard on the conversion paths.
+A faithful port would need those call sites reworked to a conversion-free
+protocol first.
+"""
+from __future__ import annotations
+
+import os
+import platform
+
+if platform.system() == "Windows":
+    import errno as _errno
+    import msvcrt as _msvcrt
+    import time as _time
+
+    LOCK_SH = 1
+    LOCK_EX = 2
+    LOCK_NB = 4
+    LOCK_UN = 8
+
+    _LOCK_BYTES = 2**30
+    # Poll interval when emulating POSIX's block-until-acquired behaviour.
+    _BLOCK_POLL_SECONDS = 0.05
+
+    def flock(fd: int, operation: int) -> None:
+        if not isinstance(fd, int):
+            fd = fd.fileno()
+        # msvcrt.locking locks bytes starting from the current file position, so
+        # we must seek to 0 to lock a consistent byte range across callers.
+        # fcntl.flock leaves the file offset untouched, however, so save the
+        # caller's offset and restore it afterwards to match POSIX semantics.
+        try:
+            saved_offset: int | None = os.lseek(fd, 0, os.SEEK_CUR)
+        except OSError:
+            saved_offset = None
+        os.lseek(fd, 0, os.SEEK_SET)
+        try:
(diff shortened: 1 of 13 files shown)


CodeAbra:
Windows daemon startup is fixed in **v2.5.0**, just released: `iai-mcp daemon install` now registers a per-user Task Scheduler task, with install / start / uninstall parity with launchd and systemd. That covers the boot failure this PR targets, so I'm closing it as resolved in the release.

Thanks for surfacing the Windows gap and for the PR — the problem you flagged is what drove the fix.

PR #75, opened 2026-07-15
Fix Windows regressions since the v2.1.0 port

LC1207:
## What this fixes

The v2.1.0 community port had Windows working end to end — thank you for taking that in. Best we can tell, a handful of Unix-only constructs crept back in between v2.2.0 and v2.3.1, and on Windows the v2.3.1 daemon now fails in two stages: it crashes at boot (`import resource`, then `signal.SIGHUP` in the shutdown-handler tuple), and once past those it runs but is unreachable — `SocketServer.serve()` inspects `inspect.signature(asyncio.start_unix_server)` before its `IS_WINDOWS` branch, and since `start_unix_server` doesn't exist on Windows the listener task dies at birth while the daemon keeps ticking. `daemon status` then reports "not running" against a live daemon and `.daemon.port` goes stale.

## Changes (one concern per commit)

1. **Daemon boot + socket serving** — guard `import resource` (fd rlimits are a Unix concern), resolve `SIGHUP` with `getattr` so the tuple itself can't raise, and move the `start_unix_server` signature inspection below the Windows early-return.
2. **File locking** — `capture_queue.py`, `lifecycle.py`, and `lifecycle_event_log.py` import `fcntl` directly; swapped to the `_flock` seam the port introduced (drop-in for `flock`/`LOCK_*`). `lillibrain/io.py` gets a guarded import since its `F_FULLFSYNC` use is already `_DARWIN`-gated.
3. **CLI + doctor** — `cmd_daemon_start` called `os.getuid()` before its OS dispatch (crash instead of the intended "Unsupported OS" message); doctor check (c) now uses the `_flock` seam, and check (b) probes `.daemon.port` on Windows instead of the nonexistent `.daemon.sock`.

## Validation

Windows 11, Python 3.12, editable install, daemon under Task Scheduler: daemon boots and serves, `daemon status` responds, `doctor` passes 26/26 (the HID-idle warn is expected there), and MCP capture/recall verified end to end from Claude Code against a store migrated from v1.2.1. The POSIX paths are untouched or moved verbatim, so this shouldn't change behavior there — but you know the codebase better; happy to split or adjust however you'd prefer.

One thought: even a minimal Windows CI job (import + boot smoke test) would catch this whole class going forward — happy to take a stab at that in a follow-up if it'd be useful.

https://claude.ai/code/session_016c4fUyixyVeVHYGrnJi98B

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -1693,7 +1693,7 @@ _PERMANENT_FAILED_NAMED_RE = re.compile(r"^(.+)\.permanent-failed-([^.]+)\.jsonl
 
 def _count_lines(fpath: Path) -> int:
     try:
-        with fpath.open() as fh:
+        with fpath.open(encoding="utf-8", errors="replace") as fh:
             return sum(1 for ln in fh if ln.strip())
     except OSError:
         return 0
@@ -1758,7 +1758,7 @@ def drain_permanent_failed_files(
         file_dropped = 0
 
         try:
-            with fpath.open() as fh:
+            with fpath.open(encoding="utf-8", errors="replace") as fh:
                 lines = [ln.rstrip("\n") for ln in fh if ln.strip()]
 
             if not lines:
--- a/src/iai_mcp/doctor/_lifecycle_checks.py
+++ b/src/iai_mcp/doctor/_lifecycle_checks.py
@@ -177,7 +177,8 @@ def check_b_socket_fresh() -> CheckResult:
 
 def check_c_lock_healthy() -> CheckResult:
     import errno as _errno
-    import fcntl as _fcntl
+
+    from iai_mcp import _flock as _fcntl  # flock seam: fcntl on POSIX, msvcrt region lock on Windows
 
     lock_path = _resolve_hippo_db_path().parent / ".lock"
     if not lock_path.exists():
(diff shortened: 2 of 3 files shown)


CodeAbra:
v2.5.0 just landed and it brought a substantial amount of Windows work (Task Scheduler daemon lifecycle, boot shims). Could you rebase this onto the current `main`? Some of what this PR fixes may already be covered, and a rebase will show what's still needed on top. I'll review the remaining delta as soon as it's rebased — thanks for staying on the Windows regressions.

LC1207:
Rebased onto current main (fe20ebc). You were right — 2.4.1/2.5.0 covered most of it: the resource guard, lazy SIGHUP, the _flock seam on every caller we'd patched (daemon_state included), the getuid branching, and doctor's check (b) PORT_FILE probe are all upstream now, so those commits were dropped.

Two survived the rebase, verified still failing on main under Python 3.12 on Windows:

- socket_server.py — inspect.signature(asyncio.start_unix_server) runs before the IS_WINDOWS early-return; the attribute doesn't exist on Windows (≤3.12), so serve() dies before it ever reaches the new IPC branch and the daemon runs unreachable.
- doctor check (c) — bare import fcntl outside the check's try block; routed through the iai_mcp._flock seam like its siblings.

The PR is now 2 commits, 2 files, +6/−4. I'm running the rebased build on a live Windows store — daemon status and doctor clean, MCP tools serving. Thanks for the Windows work in 2.5.0; daemon install registering the Task Scheduler task directly is much nicer than the hand-rolled task we'd been running.

PR #79, opened 2026-07-17
feat(daemon): add WAKE-state drain valve for the capture spool

driize:
# feat(daemon): add WAKE-state drain valve for the capture spool

## Motivation

The spool→store drain currently runs in exactly two places: daemon startup and the WAKE→DROWSY edge (`_should_drain_on_drowsy_edge`). DROWSY requires ≥300s without a heartbeat — but the MCP wrapper refreshes the heartbeat every 30s unconditionally while the editor is open. For anyone who keeps their editor open continuously, "between sessions" never arrives: captures accumulate in `.deferred-captures/` for days and nothing reaches the store until the daemon restarts. (Observed in practice: 59 spool files, store 185→318 records only after a manual restart.)

This adds an escape valve: while in WAKE, periodically probe the spool cheaply and run the existing drain when a real backlog has built up.

## Design

- `spool_backlog_snapshot()` (capture.py): `os.scandir` + name regexes only — counts finalized files (`*.crash-N.jsonl`, `*.processing-PID.jsonl`) and sums `st_size` of `*.live.jsonl`. Never opens capture content.
- In the lifecycle tick, sibling to the drowsy-edge block and modeled on the existing pending-embed floor pass:
  - `IAI_MCP_WAKE_DRAIN_FLOOR_SEC` (default 600): probe cadence; `<= 0` disables the valve entirely (not even the probe runs).
  - `IAI_MCP_WAKE_DRAIN_MIN_BYTES` (default 65536): fire when live-file bytes have grown by at least this much since the last calibration point. The baseline is a process-local latch: the first probe calibrates without firing, shrink (rotation/removal) recalibrates, and every fire advances the baseline so a drain that can't make progress doesn't refire in a loop. Any finalized (crashed/orphaned) file fires immediately.
  - Trigger decisions are pure functions (`_wake_valve_due`, `_spool_backlog_over_threshold`) with dedicated unit tests, in the same style as `_should_drain_on_drowsy_edge`.
- The drain itself is the existing `drain_capture_backlog` — already single-flighted, RSS-capped and batch-capped (`MAX_DRAIN_EVENTS_PER_RUN`), so concurrent triggers degrade to a no-op skip. Runs via `asyncio.to_thread`, same shape as the drowsy drain.
- `_run_drowsy_drain` gained optional `event_name`/`phase` parameters; defaults keep the drowsy edge byte-for-byte unchanged. The valve emits `deferred_drain_wake_valve` on success-with-work; failures keep `deferred_drain_failed` with `phase="wake_valve"`.

Startup and drowsy-edge drain behaviour is untouched; this is purely additive and can be disabled with one env var.

## Test results

Targeted: `tests/test_daemon_wake_valve.py` — 43 passed (trigger pure functions, snapshot probe, baseline latch semantics, disable knob, drowsy-edge unchanged). Full suite on this branch: 5164 passed, 228 skipped, 4 xfailed, 2 failed — both failures (`test_rss_bounded`, `test_subagent_spawns_zero_new_processes`) are environment-sensitive and already fail on a fresh `upstream/main` baseline on the same machine (3 failed, 5144 passed).

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -1158,6 +1158,46 @@ def write_deferred_captures(
     return out_path
 
 
+def spool_backlog_snapshot(deferred_dir: Path | None = None) -> dict[str, int]:
+    """Cheap, drain-free estimate of how much backlog is sitting in the spool.
+
+    Used as a trigger signal only (e.g. the WAKE drain valve) -- it never
+    opens capture content or any per-session offset file, only ``stat()``s
+    the spool directory entries, so it is safe to call frequently even under
+    heavy backlog. ``live_bytes_total`` is the raw total size of the still-
+    open ``*.live.jsonl`` files; callers that need a *delta* (growth since
+    the last probe) track their own baseline against this total -- the
+    per-session drain-offset is a line count, not a byte count, so it cannot
+    be subtracted from a byte size here. False positives are fine and
+    expected; the expensive verification happens in the drain itself.
+    """
+    if deferred_dir is None:
+        deferred_dir = Path.home() / ".iai-mcp" / ".deferred-captures"
+    snapshot = {"finalized_files": 0, "live_bytes_total": 0}
+    if not deferred_dir.exists():
+        return snapshot
+
+    try:
+        with os.scandir(deferred_dir) as it:
+            for entry in it:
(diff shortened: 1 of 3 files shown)


driize:
Withdrawn — this feature PR was opened prematurely; the two fix PRs (#77, #78) stand.

PR #84, opened 2026-07-23
Source-aware recall weighting: soft-boost curated documents and durable records

Marsu6996:
## What

Adds an optional soft multiplicative boost at the final ranking stage of recall, favouring records that carry durable knowledge over raw episodic turns:

- boosted: records tagged `doc:*` (i.e. ingested via `upload`/`teach`) and `semantic`-tier records;
- explicitly NOT boosted: auto-induced `schema` / `pattern:*` records (inference artefacts, not knowledge);
- never an eligibility filter — candidates are re-scored, never excluded;
- applied consistently on the three recall paths (main pipeline, exact-cosine authority merge, direct fallback);
- zero additional store queries: tier/tags come from the already-batched payloads;
- configurable: `IAI_MCP_SOURCE_WEIGHT_FACTOR` (default 1.05, clamped to [1.0, 1.25], 1.0 = disabled).

## Why

On a real 18k-record French store (Solon-large HTTP provider), measured through the daemon path against a 63-query human-labelled gold set, a factor sweep gave:

| factor | top-1 | top-5 | MRR | curated share of top-5 hits |
|---|---|---|---|---|
| 1.0 (off) | 12.7% | 25.4% | 0.187 | 11% |
| **1.05** | **14.3%** | **28.6%** | **0.201** | **27%** |
| 1.10 | 11.1% | 28.6% | 0.185 | 42% |

1.05 improves every column; 1.10 over-promotes curated chunks past correct top-1 records, hence the conservative default.

## Tests

Unit tests cover: boost applied to `doc:`/semantic records, neutrality at factor 1.0, no boost for `schema`/`pattern:*` semantic records, clamping and env parsing. Targeted recall suite passes (108 passed).

🤖 Designed by Marsu — Refined by Claude.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.5.1+m1] — 2026-07-23
+## [Unreleased]
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.5.1+m1] — 2026-07-23
+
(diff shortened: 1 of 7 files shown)


Marsu6996:
Closing in favour of #85 — the ranking work is still moving on our side (a hybrid lexical lane now changes the context this boost operates in), so a discussion issue with the full measurements is more honest than a point-in-time PR. Will re-propose once the design has settled. 🤖 Designed by Marsu — Refined by Claude.

Issue #86, opened 2026-07-24
[bug] minor - Unhandled UnicodeDecodeError from "pmset -g log" fails the lifecycle tick


res-pstepan:
## Summary

`idle_detector.pmset_recent_sleep` reads `pmset -g log` with `subprocess.run(..., text=True)`, which decodes the output as strict UTF-8. When that output contains a byte that is not valid UTF-8, the decode raises `UnicodeDecodeError`. The function catches only `FileNotFoundError`, `subprocess.TimeoutExpired`, and `OSError`, so the decode error is not caught and the lifecycle tick fails.

## Environment

- iai-mcp version: 2.6.0
- macOS version: 26.5.1
- Python version: 3.11.15
- Node version: v26.5.0
- MCP host: Claude Code

## The issue

`pmset -g log` on my machine is about 8.85 MB. Most of the time it decodes cleanly. The daemon logged this traceback 33 times in a row, then the output was valid again:

```
File "src/iai_mcp/idle_detector.py", line 87, in pmset_recent_sleep
    result = subprocess.run([_PMSET_BIN, "-g", "log"], capture_output=True, text=True, ...)
File ".../subprocess.py", line 1086, in _translate_newlines
    data = data.decode(encoding, errors)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd2 in position 9004805: invalid continuation byte
```

Each occurrence surfaces as `lifecycle tick iteration failed` (`lifecycle_tick` calls `sleep_eligible`, which calls `pmset_recent_sleep`).

The daemon keeps running. The affected tick loses its pmset sleep-eligibility check.

Same code path in 2.4.1 and 2.6.0.

CodeAbra:
Fixed and released in v2.6.1.

`text=True` decodes strict UTF-8, and `UnicodeDecodeError` is a `ValueError` — so it slipped past the `OSError` chain and escaped the detector. All four subprocess readers shared that pattern (`ioreg` had the same exposure), so the fix covers all four: one helper, bytes captured and decoded leniently. A log carrying both a junk byte and a real sleep event still reports the sleep — now a regression test.

Thanks — you found the class, not just the instance.

Issue #85, opened 2026-07-23
Recall quality on a real 18k store: honest measurements, early experiments, looking for the lever that moves it drastically


Marsu6996:
We are measuring and trying to drastically improve recall quality on a real production store (18k records, French, custom Solon-large HTTP embedding provider), against a 63-query human-labelled gold set evaluated through the daemon path. Opening this issue to share findings as we go and to discuss directions — this is very much work in progress.

## What we measured

- Baseline (v2.5.1, exact index): **top-1 12.7%, top-5 25.4%**. The move from ANN to the exact index changed almost nothing on this benchmark — retrieval approximation was not the bottleneck.
- ~90% of top-5 hits were raw episodic turns, even when curated `doc:*` chunks (ingested via `upload`/`teach`) containing the answer sat just below the cutoff: ranking does not distinguish knowledge from conversation.
- The vocabulary gap dominates failures: a query phrased with words absent from the stored text ("versionnage" vs a record talking about semver/release/bump) leaves the target stranded at deep ranks.

## What we tried so far (branches on our fork)

- **Source-aware weighting** ([m1-source-weighting](https://github.com/Marsu6996/iai-personal-memory-engine/tree/m1-source-weighting)): bounded soft boost of `doc:*`/semantic-tier records at final ranking, never a filter. Factor swept on the gold set; 1.05 optimal.
- **IDF-gated hybrid lexical⊕semantic** ([m2-hybrid-lexical](https://github.com/Marsu6996/iai-personal-memory-engine/tree/m2-hybrid-lexical)): extends the in-tree BM25 `LexicalIndex` with FR tokenisation + deterministic synonyms, fires only on rare tokens (IDF ≥ 4), RRF fusion; RRF-k swept, 240 optimal.

**Honest outcome: these two combined only moved top-1 from 12.7% to 14.3%** (top-5 did better: 25.4% → 34.9%, and six previously unfindable queries now surface). That is far from what we consider acceptable — our target is near-assured top-5 recall.

## Where we are heading

We are actively working on this and looking for the lever that moves the score drastically, not incrementally. Candidates on our list: a much larger gold set to understand the remaining failures, chunking/section-aware ingestion, and a measured reranker. **We are open to any and all ideas** — engine-side, ingestion-side, evaluation-side, or approaches we have not considered at all. If you have thoughts on where the engine itself could help most, or results from your own stores, we are very interested.

🤖 Designed by Marsu — Refined by Claude.

[...]

PR #89, opened 2026-07-25
Add decoding regression coverage for all idle-detector readers

Marsu6996:
## Summary

Repointed against current `main` as requested after the production fix shipped in v2.6.1. This PR now contains regression coverage only; it makes no production-code changes.

The shared lenient-decoding helper serves four system readers. The test coverage now exercises all four paths:

- `ioreg` HID idle output (existing invalid-UTF-8 test retained);
- `pmset -g log`, strengthened so the invalid byte occurs inside an otherwise valid sleep-event line;
- `pmset -g` responsiveness output;
- Linux `busctl` JSON output.

This is regression coverage for #86.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / tooling
- [x] Other: regression tests

## Affected areas

- [ ] Capture path
- [ ] Recall / retrieval
- [x] Daemon lifecycle / FSM
- [ ] Bench harness
- [x] Other: macOS and Linux idle detection

## Testing

- [ ] `pytest` passes locally

Targeted validation:

- `pytest -q tests/test_idle_detector.py`: 37 passed
- `ruff check tests/test_idle_detector.py`: passed
- `git diff --check`: passed

Full local suite:

- 5511 passed, 216 skipped, 4 xfailed, 2 failed

Both failures are checkout-environment mismatches unrelated to this test-only diff:

- the local installed distribution is still 2.6.0 while current `main` declares 2.6.1;
- the checkout contains both Python 3.11 and Python 3.12 native extension binaries, while the test requires exactly one.

The pull-request CI will validate the branch in a clean environment. The repository-wide Ruff baseline is not currently clean; the touched file introduces no Ruff diagnostics.

## Benchmarks

Not applicable: this PR changes tests only.

## Notes for reviewers

The previous implementation change has been removed because v2.6.1 already ships the shared `_run_captured` fix. The branch is rebased onto current `main` and now preserves only the stronger regression coverage requested in review.

### DIFF
--- a/tests/test_idle_detector.py
+++ b/tests/test_idle_detector.py
@@ -7,7 +7,7 @@ from datetime import datetime, timedelta, timezone
 from unittest.mock import patch
 
 
-from iai_mcp.idle_detector import IdleDetector, IdleStatus
+from iai_mcp.idle_detector import IdleDetector, IdleStatus, _pmset_responsive
 
 
 def _completed_process(
@@ -113,10 +113,10 @@ def test_pmset_recent_sleep_returns_false_when_pmset_missing() -> None:
 
 
 def test_pmset_recent_sleep_survives_invalid_utf8_in_log() -> None:
-    log = _pmset_log_stdout([
-        (_now_pmset_ts(offset_min=2), "System Sleep"),
-    ])
-    corrupted = b"\xd2 garbage line\n" + log.encode("utf-8")
+    corrupted = (
+        f"{_now_pmset_ts(offset_min=2)} ".encode()
+        + b"\xd2 System Sleep Notification clientId=foo\n"
+    )
     fake = _completed_process(stdout=corrupted)
     with patch("iai_mcp.idle_detector.subprocess.run", return_value=fake):
         result = IdleDetector().pmset_recent_sleep(window_min=5)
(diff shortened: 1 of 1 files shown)


CodeAbra:
#88 is merged — thanks.

This one collided. The same fix went out in v2.6.1 a few minutes ago; it was written and gate-tested before your PR opened, and you had no way to see it — our work happens in a private tree first. Ours routes all four readers through one lenient-decoding helper, including the Linux `busctl` probe you scoped out.

Your tests are better than what shipped, though. Ours assert end behaviour; your fake `run` asserts that `encoding` and `errors` are actually passed, which catches someone quietly dropping them later. If you repoint this PR at just those tests against current `main`, I'll take it.

CodeAbra:
Closing this one — the fix it covers went out in v2.6.1 back in July, written and gate-tested before this PR opened, so there is nothing left for it to land.

Your reading of the bug was right, and I said so at the time: `text=True` decodes strict UTF-8 and `UnicodeDecodeError` is a `ValueError`, so it slipped past the `OSError` chain entirely. All four subprocess readers now decode with `errors="replace"`.

3.0.0 is out today. Two things in it are aimed at what you reported in #85, and I'll write there separately rather than bury it here.

PR #93, opened 2026-08-05
Resolve the iai-mcp CLI from PATH in the session-recall hook

ZhdanDesign:
## Summary

`iai-mcp-session-recall.sh` looked for the CLI in only two hardcoded locations (`~/IAI-MCP/.venv/bin`, `/usr/local/bin`). A standard `pip install` puts it in `~/.local/bin`, and pipx/pyenv/Homebrew installs land elsewhere again, so the hook found nothing.

Because the hook is fail-safe by design (exit 0, empty stdout on every error path), the result is silent: session recall is simply never injected, and the only trace is a `skipped: iai-mcp CLI not found` line in `~/.iai-mcp/logs/recall-YYYY-MM-DD.log`.

`iai-mcp-session-capture.sh` already solves this — it consults `command -v iai-mcp` before its candidate list, and that list covers pyenv shims, pipx, Homebrew and `~/.local/bin`. This PR gives the recall hook the same resolution order, so the two halves of the hook pair agree.

Found on a macOS install where capture worked and recall silently did not.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Build / tooling

## Affected areas

- [ ] Capture path
- [x] Recall / retrieval
- [ ] Daemon lifecycle / FSM
- [ ] Bench harness
- [ ] CLI / doctor
- [x] Other: Claude Code hook scripts

## Testing

- [x] `pytest` passes locally
- [x] New tests added for changed behaviour

`tests/test_session_recall_hook_cli_lookup.py` covers three cases: resolution via PATH, resolution via `~/.local/bin` when PATH has no entry, and the unchanged fail-safe when no CLI exists anywhere.

Verified red/green against the unpatched hook:

```
# origin/main
FAILED test_hook_resolves_cli_from_path
FAILED test_hook_resolves_cli_from_user_local_bin
2 failed, 1 passed

# this branch
3 passed
```

`tests/test_plugin_manifest.py` passes, so the `plugin/hooks/` and `src/iai_mcp/_deploy/hooks/` copies remain byte-identical.

I could not run `ruff` — it is not installed in my environment, and I did not want to add it to the venv this box runs the daemon from. The change is shell-only plus one new test file written in the style of `tests/test_session_recall_hook_failsafe_exit_zero.py`.

## Benchmarks

Not applicable — no retrieval, capture, or consolidation logic changes. The patch only affects how the hook locates the CLI binary.

## Notes for reviewers

- No behaviour change when the CLI is already found: the env-var override and `.cli-path` cache still take priority, in that order.
- The fail-safe contract is preserved — the added lookup cannot introduce a non-zero exit.
- `iai-mcp-per-turn-recall.sh` and `iai-mcp-turn-capture.sh` do not invoke the CLI, so they need no equivalent change.

### DIFF
--- a/plugin/hooks/iai-mcp-session-recall.sh
+++ b/plugin/hooks/iai-mcp-session-recall.sh
@@ -73,8 +73,11 @@ fi
 #   1. IAI_MCP_SESSION_RECALL_CLI environment variable (developer override
 #      for non-standard install locations; export in your shell init).
 #   2. ~/.iai-mcp/.cli-path cache file (auto-populated below once the
-#      candidates array finds a working binary).
-#   3. Generic install locations in the candidates array.
+#      lookup finds a working binary).
+#   3. `command -v iai-mcp` — PATH lookup; picks up pyenv shims, pipx
+#      wrappers, and any other PATH-managed install transparently.
+#   4. Generic install locations in the candidates array, checked when PATH
+#      has no entry.
 # Only generic install paths are baked into the source; developer-specific
 # paths belong in the env var or the cache, never here.
 cli_cache="$HOME/.iai-mcp/.cli-path"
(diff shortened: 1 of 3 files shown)


CodeAbra:
Merged, thanks. Capture worked and recall didn't, and the fail-safe exit meant nobody would notice. That could have sat there for months. Your red/green run against origin/main made it quick to check.

PR #95, opened 2026-08-06
feat(windows): Support PowerShell hooks and GUI Desktop auto-scan for Antigravity

owlze:
Fixes Windows hook generation for Antigravity.

### Changes:
1. Translated bash templates to .ps1 for Windows.
2. _antigravity_hooks.py dynamically writes powershell.exe into hooks.json if os.name == 'nt'.
3. Automates the background log ingestion for Antigravity Desktop (GUI) users via Windows Task Scheduler (schtasks) because the GUI does not trigger hooks.json.

Thank you!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,8 @@ iai_mcp = [
     "_deploy/systemd/*.service",
     "_deploy/windows/*.xml",
     "_deploy/hooks/*.sh",
+    "_deploy/hooks/*.ps1",
+    "_deploy/windows/*.xml",
     "_deploy/brainview/*.html",
     "_wrapper/*.js",
 ]
--- a/src/iai_mcp/_deploy/hooks/iai-mcp-antigravity-capture.ps1
+++ b/src/iai_mcp/_deploy/hooks/iai-mcp-antigravity-capture.ps1
@@ -32,3 +32,4 @@ if (Test-Path $liveFile) {
     $newName = "$sessionId.live-$ts-$pidId.jsonl"
     Rename-Item -Path $liveFile -NewName $newName -ErrorAction SilentlyContinue
 }
+
--- a/src/iai_mcp/_deploy/hooks/iai-mcp-antigravity-capture.ps1
+++ b/src/iai_mcp/_deploy/hooks/iai-mcp-antigravity-capture.ps1
@@ -1,4 +1,4 @@
-﻿$inputJson = $input | Out-String | ConvertFrom-Json -ErrorAction SilentlyContinue
+$inputJson = $input | Out-String | ConvertFrom-Json -ErrorAction SilentlyContinue
 if (-not $inputJson) { exit 0 }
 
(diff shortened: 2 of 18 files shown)


CodeAbra:
Three of the four asks landed — package data, the XML task template, the uninstall delete, and the test. Thanks.

Two things left.

**The hardcoded CLI path is still there.** Both `.ps1` hooks add:

```
$cli = "C:\iai-personal-memory-engine\.venv\Scripts\iai-mcp.exe"
```

On any other install that check fails and the hook exits 0 silently. Same class we fixed in #93 for the bash hooks. This one needs you: it is your file and you have a Windows box, we develop on mac only and cannot test it.

**Your editor rewrote our files as cp1251.** 41 em-dashes became `вЂ”` and 10 files gained a BOM, across `lifecycle_lock.py`, `fsm_reconcile.py`, `daemon_state.py`, `lifecycle_state.py`, `user_model.py`, `migrate/_reembed.py` and `provenance_buffer.py`. Nothing functional, but it lands in comments that document lock ordering and staging invariants. Set the editor to UTF-8 without BOM before the next push.

Don't spend time reverting that by hand — I will clean it in a commit on top after the merge, and add a guard test so it cannot happen again.

Keep everything else as it is. The 17 `encoding="utf-8"` additions, the `fcntl` guard and the `os.fchmod` guard are real portability bugs and the most valuable part of this PR.

CodeAbra:
Merged as b9f3c10 — thanks. The Windows portability work is the part I value most: 17 write sites now carry `encoding="utf-8"`, `fcntl` degrades instead of crashing, and `os.fchmod` is guarded. Those were real bugs on your platform that we could never have found on ours.

Two follow-ups landed on top, so you know what changed and why.

**c9c3483** restored the em-dashes and dropped the byte-order marks in the lifecycle, FSM, state and migration modules, and removed the debug prints from `cmd_session_start` and `cmd_capture_turn_deferred` — those write into the hook streams on every session.

It also reworked `tests/test_antigravity_hooks_windows.py`. Your version reloaded the module under a patched `platform.system` and never reloaded it back, so `_EXT` stayed `.ps1` for the rest of the session and two existing installer tests failed on macOS CI:

```
ERROR: hook templates missing in package data:
  ['iai-mcp-session-capture.ps1', 'iai-mcp-session-recall.ps1', 'iai-mcp-per-turn-recall.ps1']
```

The installer was right; the test leaked. It now restores the module in an autouse fixture teardown, covers Linux, Darwin and Windows, and asserts `_CMD_PREFIX` too. Reproduced red with your version and green with the new one on the same tree.

**6410731** adds a guard test so neither problem can return: it scans every tracked text file for cp1251 mojibake, for byte-order marks outside `.ps1`, and for debug prints under `src/`. Your `.ps1` marker is deliberately allowed — PowerShell 5.1 needs it.

If you want to keep working on Windows support, the open piece is that `cursor`, `hermes`, `antigravity` and `codex` capture adapters do not extract tool blocks yet.

PR #96, opened 2026-08-08
feat(capture): extract tool blocks from adapter transcripts

owlze:
?? Hi again! Got inspired by the idea of extracting tool blocks and implemented it for Antigravity, Cursor, Codex and Claude.
Tested on live Antigravity transcripts  works perfectly!

This PR extracts the `tool_calls` and formats them as markdown blocks so the underlying commands/edits are preserved in the memory engine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,8 @@ iai_mcp = [
     "_deploy/systemd/*.service",
     "_deploy/windows/*.xml",
     "_deploy/hooks/*.sh",
+    "_deploy/hooks/*.ps1",
+    "_deploy/windows/*.xml",
     "_deploy/brainview/*.html",
     "_wrapper/*.js",
 ]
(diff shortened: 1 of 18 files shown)


CodeAbra:
Thanks for picking this up so fast, and sorry — I'm going to close this one, and I want to be straight about why, because it isn't a problem with your code.

Tool extraction has to satisfy an invariant that lives in the engine internals: **all three capture carriers must emit byte-identical text.** There is the batch `capture_transcript` path, the deferred CLI path, and the Stop-hook shell parser, and if any two of them format a turn differently, hosts without stable turn uuids mint duplicate records. There's also a trap our own review hit twice: in real transcripts `tool_result` entries carry role `user`, so an aggregation that resets pending tools on user-role entries records no mechanics at all on the dominant path.

Neither of those is discoverable from outside the engine. That's on us for not writing it down before pointing you at the task — the follow-up note I left you was one line, and the constraint behind it is three paragraphs.

So the adapter path is going to the engine owner, who is doing all four adapters in one pass with the carrier parity tests. Your version won't be merged and I'd rather tell you today than ask you to rebase first and then say it.

One more thing worth naming: this PR branched from your pre-#95 work, so it re-carries the whole #95 file set on top of a main that already has it. Even without the above it would have needed a rebase down to just the `capture.py` change.

**What is genuinely yours, if you want it:** Windows. The runtime is ported but the test suite is not, and we develop on macOS only — we cannot verify a single Windows path ourselves. Every test you get running there is something we physically cannot do. `iai-mcp doctor` on Windows is a good place to start: it has 27 checks and several of them assume POSIX.

3.0.0 goes out today with your PowerShell hooks and your portability fixes in it. The encoding work turned out to be the seed of a wider sweep — 63 declared sites became 161 across the tree.

PR #97, opened 2026-08-09
fix(windows): resolve posix assumptions in doctor checks

owlze:
This PR fixes [WARN] (q) and [WARN] (n) false positives when running iai-mcp doctor on Windows.

- Implements a native Windows fallback using GetLastInputInfo (via ctypes.windll.user32) for HID idle time detection.
- Updates the CLI reachability check to find iai.exe when it's not directly in the system PATH but located next to sys.executable.

### DIFF
--- a/src/iai_mcp/doctor/_lifecycle_checks.py
+++ b/src/iai_mcp/doctor/_lifecycle_checks.py
@@ -707,8 +707,17 @@ def check_o_subscription_credentials() -> CheckResult:
 
 def check_q_iai_cli_reachable() -> CheckResult:
     import shutil
+    import os
+    import sys
 
     iai_path = shutil.which("iai")
+    if iai_path is None:
+        bin_dir = os.path.dirname(sys.executable)
+        exe_name = "iai.exe" if sys.platform == "win32" else "iai"
+        fallback = os.path.join(bin_dir, exe_name)
+        if os.path.exists(fallback):
+            iai_path = fallback
+
     if iai_path is None:
         return CheckResult(
             name="(q) iai CLI reachable",
--- a/src/iai_mcp/idle_detector.py
+++ b/src/iai_mcp/idle_detector.py
@@ -207,6 +207,27 @@ class IdleDetector:
         return min(idle_times)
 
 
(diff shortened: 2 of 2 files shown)


CodeAbra:
Windows part reads right to me — the `iai.exe` fallback next to `sys.executable` and the `GetLastInputInfo` probe with the tick-count wraparound both look correct. Windows stays yours.

One thing to put back before I merge. The logind branch got rewritten too, and it isn't part of the Windows fix:

```python
        if "logind" in status.available_signals:
            detail = (
                f"logind IdleHint: {idle_str('not idle')}, available: {signals_str}"
            )
            return detail, "PASS"
```

`tests/test_doctor.py::test_doctor_row_n_hid_idle_source_linux_logind` asserts that exact string — I ran your branch on mac and got `1 failed, 51 passed` on it. And `not idle` is the accurate label in that branch: logind answered and the user is active, so `unavailable` says the opposite of what happened.

Your new Windows branch below it is fine as written, no change needed there.

Small thing while you're in the file: lines 214 and 220 have trailing whitespace.

Note on platform: we develop on mac, so nothing Windows here gets verified on our side — your test on a real Windows box is the check that counts.

CodeAbra:
Hi @owlze — just a nudge on this one. The Windows fixes look good; the only thing holding it is the logind line in `idle_detector.py` that got rewritten and breaks `test_doctor.py::test_doctor_row_n_hid_idle_source_linux_logind`. Could you revert just that line? Happy to merge once that's back.

CodeAbra:
Hey @owlze, circling back one last time. Your Windows doctor fixes are good and I'd like to land them. The only holdout is still that one logind line in `idle_detector.py` that breaks `test_doctor_row_n_hid_idle_source_linux_logind`. If a round-trip is a hassle, just say the word and I'll revert that single line on my end and merge under your name. Or push the revert yourself, whichever's easier. Either way I'd like to get this in.

CodeAbra:
This one's independent of the daemon rework and still merges cleanly. CI is red/unstable though — could you rebase onto current main and check the failing job? Once it's green I'll take it.

PR #102, opened 2026-08-11
Fix FileNotFoundError race in turn-capture offset writes under concurrency

### LINKED ISSUE #98: Concurrent turn-capture invocations race on a shared tmp path, raising FileNotFoundError
**What happens**

UserPromptSubmit's turn-capture hook and Stop's session-capture hook (which calls `capture-turn-deferred`) both maintain the same `{session_id}.offset` file, and can fire close together for one exchange — confirmed live on Cursor, which fires the turn-capture hook twice per prompt. Both writers build their rename-source tmp path from the session id alone, with no per-process uniqueness:

```python
tmp_path = offset_path.parent / (offset_path.name + ".tmp")
```

Whichever process calls `os.replace()` second finds its tmp file already consumed by the first, raising `FileNotFoundError`. The exception is caught and logged, never fatal — but the write it belonged to is lost. On a session's first turn (no prior offset baseline yet) this can drop the write entirely until the Stop-hook safety net catches up.

Same pattern exists at three call sites sharing this shape: the offset write, and the `watermark`/`live-fingerprint` tmp writes used by the per-turn freshness gate.

**Reproduced empirically**

A 20-process concurrency stress test against `capture-turn-deferred` (10 pairs of `iai-mcp-turn-capture.sh` + `iai-mcp capture-turn-deferred`, launched simultaneously against the same session) failed 9/20 calls with this exact `FileNotFoundError`.

**Fix**

PR incoming: suffix every tmp path with `os.getpid()`, so concurrent writers never share a rename source. Same stress test passes 20/20 after the fix, with no leftover tmp files and the correct final offset.

tom-shields-stitch:
Fixes #98.

UserPromptSubmit's turn-capture hook and Stop's session-capture hook (which calls `capture-turn-deferred`) both maintain the same `{session_id}.offset` file, and can fire close together for one exchange — confirmed live on Cursor, which fires the turn-capture hook twice per prompt. Both writers built their rename-source tmp path from the session id alone, no per-process uniqueness, so whichever process called `os.replace()` second found its source already consumed by the first, raising `FileNotFoundError`. The error is caught and logged, never fatal, but the write it belonged to is lost — on a session's first turn (no prior offset baseline) this can drop the write entirely until the Stop-hook safety net catches up.

## Fix

Suffix every tmp path with `os.getpid()`, so concurrent writers never share a rename source. Applies to the three call sites sharing this pattern (offset, watermark, live-fingerprint tmp writes) in both the turn-capture hook script and `cli/_capture.py`'s `capture-turn-deferred`.

## Verification

Not just inspection — a 20-process concurrency stress test (10 pairs of the turn-capture hook + `capture-turn-deferred`, launched simultaneously against the same session) against the unpatched code failed 9/20 calls with this exact `FileNotFoundError`. The same test against the patched code passed 20/20, with no leftover tmp files and the correct final offset.

Regression test added (`test_concurrent_calls_do_not_race_on_shared_tmp_path`) using real subprocesses via `multiprocessing`, not threads — the fix's uniqueness keys off `os.getpid()`, which threads would share, so a thread-based test would not actually exercise the fix.

## Test plan

- [x] `pytest tests/test_capture_turn_deferred.py` — 8 passed, including the new regression test
- [x] Manual concurrency stress test against both unpatched (9/20 fail) and patched (20/20 pass) code
- [x] `ruff check --select F,E9` on changed files — no new issues (pre-existing file has unrelated lint backlog per `extended-checks.yml`, which is explicitly non-blocking)

### DIFF
--- a/plugin/hooks/iai-mcp-turn-capture.sh
+++ b/plugin/hooks/iai-mcp-turn-capture.sh
@@ -115,7 +115,12 @@ if prev > total:
     # Transcript is shorter than the stored offset — it was rotated or
     # replaced.  Preserve the existing offset and skip capture to avoid
     # re-emitting old turns or clobbering a valid large offset.
-    tmp = state_dir / f"{session_id}.offset.tmp"
+    # PID-suffixed tmp name: this hook and the Stop hook capture-turn-
+    # deferred call both touch {session_id}.offset and can fire close
+    # together for one exchange. A shared tmp path lets whichever
+    # process replaces second find its source already consumed by the
+    # first (FileNotFoundError).
+    tmp = state_dir / f"{session_id}.offset.{os.getpid()}.tmp"
     tmp.write_text(str(prev))
     os.replace(tmp, offset)
     sys.exit(0)
@@ -276,7 +281,8 @@ if total > prev:
             emitted += 1
 
 new_offset = prev + consumed
-tmp = state_dir / f"{session_id}.offset.tmp"
+# PID-suffixed tmp name — see the rotated-transcript branch above for why.
+tmp = state_dir / f"{session_id}.offset.{os.getpid()}.tmp"
 with open(tmp, "w") as _f:
     _f.write(str(new_offset))
     _f.flush()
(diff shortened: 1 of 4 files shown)


CodeAbra:
Thank you for this one, and especially for the stress numbers. 9/20 failures against unpatched code is the kind of evidence we can act on immediately.

It turns out the engine already carries this fix, and covers one seam yours doesn't: turnstate, four write paths in total. So I'm closing it as already-fixed rather than as wrong.

Your 20-process subprocess shape is going into our own pin. You're right that threads share the pid and wouldn't exercise the fix, and our test had that gap. Credited in the changelog.

PR #103, opened 2026-08-11
Wire should_bootstrap_trigger into the consolidation gate

### LINKED ISSUE #99: should_bootstrap_trigger is defined and imported but never called
**What happens**

`quiet_window.py` has a complete, already-tested bootstrap fallback: `should_bootstrap_trigger()` returns `True` after `BOOTSTRAP_IDLE_HOURS` (2h) of inactivity, meant to cover a fresh install for the `MIN_DAYS_FOR_LEARN` (7-day) window before the presence-based quiet-window learner has enough history to run.

It's imported into `daemon/__init__.py`:

```python
from iai_mcp.quiet_window import (
    ...
    should_bootstrap_trigger,
    ...
)
```

but grepping the whole package for `should_bootstrap_trigger(` turns up only its own definition — nothing ever calls it.

**Effect**

A fresh install's automatic nightly consolidation depends entirely on the fixed night-default window (2-6am local, from `effective_consolidation_window`'s fallback) landing on a stretch where the daemon happens to be running continuously with genuine idle time — with no faster path available during the whole first week. In practice (verified on a real 5-day-old install with heavy multi-session usage) this meant zero automatic consolidation had ever run: `lifecycle status` showed `WAKE` for 5 straight days across multiple daemon restarts, and `events_query` for trajectory events returned empty.

**Fix**

PR incoming: wire the check into `_in_consolidation_window` (the gate `_may_enter_sleep` actually consults) — before a quiet window has been learned, check `should_bootstrap_trigger` against the lifecycle state's `last_activity_ts` before falling through to the fixed-window check.

tom-shields-stitch:
Fixes #99.

`quiet_window.py` has a complete, already-tested bootstrap fallback: `should_bootstrap_trigger()` returns `True` after `BOOTSTRAP_IDLE_HOURS` (2h) of inactivity, meant to cover a fresh install for the `MIN_DAYS_FOR_LEARN` (7-day) window before the presence-based quiet-window learner has enough history to run. It was imported into `daemon/__init__.py` but never actually called anywhere — the only two matches for the name were its own definition and that unused import.

In practice this meant a fresh install's automatic nightly consolidation depended entirely on the fixed night-default window (2-6am local, from `effective_consolidation_window`'s fallback) landing on a stretch where the daemon happened to be running continuously with genuine idle time — with no faster path available during the whole first week.

## Fix

`_in_consolidation_window` (the actual gate `_may_enter_sleep` consults) now checks `should_bootstrap_trigger` against the lifecycle state's `last_activity_ts` whenever no `quiet_window` has been learned yet, before falling through to the fixed-window check. Contained to this one function — no changes to `_may_enter_sleep`'s signature or callers — consistent with its existing fail-open contract (any error here still allows consolidation rather than blocking it).

## Verification

Existing `test_should_bootstrap_trigger_2h_idle` already covered the pure function in isolation and continues to pass unmodified. Added `test_bootstrap_trigger_actually_reaches_consolidation_gate`, which exercises the real gate the daemon consults (not the pure function alone), against an isolated store — confirms the gate returns `True` after 2h idle pre-learning, and is correctly skipped once a window exists.

Also verified this against a real install that had been running 5 days with heavy multi-session usage: `lifecycle status` showed `WAKE` continuously across multiple daemon restarts, and `events_query` for trajectory events was empty — consistent with this trigger never having fired.

## Test plan

- [x] `pytest tests/test_quiet_window.py` — 9 passed, including the new integration test
- [x] Manual isolated-store test confirming the gate now returns `True` after 2h idle with no learned window, and is unaffected once a window exists

### DIFF
--- a/src/iai_mcp/daemon/__init__.py
+++ b/src/iai_mcp/daemon/__init__.py
@@ -504,6 +504,29 @@ def _in_consolidation_window(daemon_state: "dict | None") -> bool:
                         return True
                 except (ValueError, TypeError):
                     pass
+        # Bootstrap fallback: before a quiet window has been learned (needs
+        # MIN_DAYS_FOR_LEARN days of history), the fixed night default still
+        # applies below, but a fresh install can go that many days without
+        # ever landing in it. should_bootstrap_trigger allows consolidation
+        # after BOOTSTRAP_IDLE_HOURS of inactivity in the meantime, regardless
+        # of time of day — same fail-open contract as the rest of this gate.
+        if daemon_state.get("quiet_window") is None:
+            try:
+                from iai_mcp.lifecycle_state import (
+                    lifecycle_state_path as _boot_lc_path,
+                    load_state as _boot_lc_load,
+                )
+                _lc_rec = _boot_lc_load(_boot_lc_path())
+                _last_activity_raw = (_lc_rec or {}).get("last_activity_ts")
+                _last_activity_dt = None
+                if _last_activity_raw:
+                    _last_activity_dt = datetime.fromisoformat(str(_last_activity_raw))
+                    if _last_activity_dt.tzinfo is None:
+                        _last_activity_dt = _last_activity_dt.replace(tzinfo=_tz.utc)
+                if should_bootstrap_trigger(_last_activity_dt, now_utc):
+                    return True
+            except Exception:
+                log.debug("bootstrap consolidation trigger check failed", exc_info=True)
         window = effective_consolidation_window(
             daemon_state.get("quiet_window"),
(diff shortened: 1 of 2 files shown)


CodeAbra:
Thank you, and I confirmed it before replying: the definition and one unused import are the only two references in the tree. It was meant to be wired from the start and never was.

The function was restructured yesterday, so your diff is written against a shape that no longer exists and can't land as-is. The check belongs inside `_window_membership_core`, before the fixed-window fallback, where both failure contracts inherit it. That's going in on our side.

Closing the PR and keeping #99 open until the fix ships, so the trail stays visible.

PR #104, opened 2026-08-11
Wire the session_exit RPC to an actual caller

### LINKED ISSUE #100: session_exit RPC handler exists but nothing ever sends the request
**What happens**

`core/__init__.py`'s dispatcher has a `session_exit` handler that calls `run_light_consolidation` (an FSRS decay/reinforcement tick over records touched in the last hour):

```python
if method == "session_exit":
    ...
    result = run_light_consolidation(store, session_id=sid)
```

Nothing in the package ever sends a `session_exit` request — no hook, no CLI command, no wrapper code. It's the light, frequent counterpart to the nightly heavy pass, and per the Stop hook's own docstring ("Claude Code fires Stop after EVERY assistant response... this hook therefore does two cheap incremental things per response"), this is exactly the kind of per-response work that hook is already designed to carry — it just never got a step added for this specific RPC.

**Effect**

Verified against a live daemon: before wiring a caller in, `select count(*) from events where kind='cls_consolidation_run'` never showed a `mode: "light"` entry from any real session, regardless of how many sessions ran.

**Fix**

PR incoming: add `cmd_session_exit` to `cli/_capture.py` (mirrors the existing `cmd_session_start`), register it as `iai-mcp session-exit --session-id ...`, and call it as a third step from `iai-mcp-session-capture.sh` (the Stop hook, shared across Claude Code, Codex and Cursor), after the existing capture-turn-deferred catch-up and live-file rotation.

tom-shields-stitch:
Fixes #100.

`core/__init__.py`'s dispatcher has a `session_exit` handler that calls `run_light_consolidation` (an FSRS decay/reinforcement tick over records touched in the last hour) — but nothing in the package ever sent a `session_exit` request. It's the light, frequent counterpart to the nightly heavy pass, and per the Stop hook's own docstring ("Claude Code fires Stop after EVERY assistant response... this hook therefore does two cheap incremental things per response"), this is exactly the kind of per-response work that hook is already designed to carry — it just never got a third step added for this specific RPC.

## Fix

- New `cmd_session_exit` in `cli/_capture.py`, following the same pattern as the existing `cmd_session_start`: sends the RPC with its own short timeouts (3s connect / 15s read) and fails safe on any error, matching the fail-safe contract of every other hook backend in this file.
- Registered as `iai-mcp session-exit --session-id ...`.
- `iai-mcp-session-capture.sh` (the Stop hook, shared across Claude Code, Codex and Cursor) now calls it as a third step, after the existing capture-turn-deferred catch-up and live-file rotation.

## Verification

Verified against a live daemon: before the fix, zero `cls_consolidation_run` events existed with `mode: "light"` from any real session. After wiring this in and calling the new CLI command directly, a genuine event appeared with the correct `session_id` and `mode`, confirming the full hook -> CLI -> RPC -> `run_light_consolidation` -> `write_event` path.

## Test plan

- [x] New `test_cmd_session_exit_daemon_unreachable.py` — covers the fail-safe contract (unreachable daemon -> `rc == 0`) and the argparse registration, mirroring the existing `test_cmd_session_start_daemon_unreachable.py`
- [x] Manual round-trip against a live daemon: `iai-mcp session-exit --session-id <test>` produced a real `cls_consolidation_run` event with the correct session_id and `mode: "light"`

### DIFF
--- a/plugin/hooks/iai-mcp-session-capture.sh
+++ b/plugin/hooks/iai-mcp-session-capture.sh
@@ -167,8 +167,16 @@ if [ -n "$session_id" ]; then
   fi
 fi
 
+# THEN the light per-response consolidation tick: FSRS decay/reinforcement
+# for records touched in the last hour (run_light_consolidation via the
+# session_exit RPC). Cheap, daemon-side, best-effort — the CLI command has
+# its own internal 3s/15s timeouts, no external `timeout` wrapper needed.
+exit_result=$("$iai_cli" session-exit --session-id "$session_id" 2>&1)
+exit_rc=$?
+
 {
   echo "$ts rc=$rc result=$result"
+  echo "$ts session-exit rc=$exit_rc result=$exit_result"
 } >> "$log" 2>/dev/null
 
 exit 0
--- a/src/iai_mcp/_deploy/hooks/iai-mcp-session-capture.sh
+++ b/src/iai_mcp/_deploy/hooks/iai-mcp-session-capture.sh
@@ -167,8 +167,16 @@ if [ -n "$session_id" ]; then
   fi
 fi
 
+# THEN the light per-response consolidation tick: FSRS decay/reinforcement
+# for records touched in the last hour (run_light_consolidation via the
+# session_exit RPC). Cheap, daemon-side, best-effort — the CLI command has
+# its own internal 3s/15s timeouts, no external `timeout` wrapper needed.
+exit_result=$("$iai_cli" session-exit --session-id "$session_id" 2>&1)
+exit_rc=$?
+
 {
   echo "$ts rc=$rc result=$result"
+  echo "$ts session-exit rc=$exit_rc result=$exit_result"
 } >> "$log" 2>/dev/null
 
 exit 0
--- a/src/iai_mcp/cli/__init__.py
+++ b/src/iai_mcp/cli/__init__.py
@@ -309,6 +309,7 @@ from ._capture import (
     cmd_session_refresh_if_stale,
     cmd_capture_transcript,
     cmd_capture_turn_deferred,
+    cmd_session_exit,
     _capture_hook_paths,
     _turn_hook_paths,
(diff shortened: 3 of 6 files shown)


CodeAbra:
Thank you for tracing this all the way to the missing caller, and for verifying it against a live daemon rather than by reading. The wiring is clean and the fail-safe matches `session_start`.

One problem with the placement. The Stop hook fires after every assistant response, not at session end, so a 100-turn session means 100 light consolidations, each recomputing FSRS over the same last-hour records. Repeated reinforcement over the same rows moves the stability curves, so "harmless" doesn't hold at that rate.

Two ways out: call it where a session actually ends, or dedupe in time, once per N minutes per session. Either works for us, and rough cost numbers would help. Happy to merge once that's settled.

tom-shields-stitch:
Good catch — you're right, and the fix was straightforward once you named the actual problem. Pushed a new commit:

Went with dedupe-in-time, not detecting a real session boundary — the RPC handler now tracks the last light-pass timestamp per session in a sidecar and skips (returning `deduped: true` + remaining cooldown) if it ran within the last 5 minutes for that session. Picked 5 minutes to match the wrapper's own idle threshold for the WAKE → DROWSY edge elsewhere in this project, so it's not an arbitrary number.

Cost numbers you asked for: confirmed against a live daemon that two immediate `session_exit` calls for the same session now produce exactly one `cls_consolidation_run` event instead of two, with the second call's response coming back as `{"mode": "light", "deduped": true, "cooldown_remaining_sec": 266.3}`. Two new tests cover the dedup itself and that it's per-session (a different `session_id` is never blocked by another session's recent pass).

CodeAbra:
Closing this one. `session_exit` is being removed as a dead path — nothing in the wrapper, the hooks, or the CLI ever sent that request, so the handler it wires up is going away rather than getting a caller. This isn't about your code; the target simply no longer exists. The dead-path finds from your reports are being fixed right now and you'll be credited for them.

PR #105, opened 2026-08-11
Wire the camouflaging_relaxation collection and weekly pass

### LINKED ISSUE #101: camouflaging_relaxation (AUTIST-13) pipeline is fully built and tested but never called
**What happens**

`iai_mcp.camouflaging` has a complete detection pipeline for the `camouflaging_relaxation` procedural knob:

- `record_user_formality(store, text, lang)` scores a turn's formality and logs a `formality_score_weekly` event.
- `run_weekly_pass(store)` aggregates that event stream and relaxes the register on a detected over-formal trend.

Every pure function here has thorough unit coverage in `tests/lilli/test_camouflaging_detection.py` — but neither end had a caller anywhere in the package:

- `capture_turn` never called `record_user_formality`, so the event stream `run_weekly_pass` depends on stayed permanently empty.
- Nothing called `run_weekly_pass` itself, so even manually-seeded data would never have been aggregated.

**Effect**

On a real install with 33k+ episodic records across 5 days of heavy multi-host usage, `profile_get_set(operation="get")` showed all 11 procedural knobs sitting at their exact seed defaults — zero drift, confirmed by diffing live values against `knobs.py`'s `default_state()` line by line.

**Fix**

PR incoming:
- `capture_turn` now calls `record_user_formality` for live, non-replayed user turns (same gate the existing foresight-refresh call already uses, so a historical backfill can't skew weeks that never happened live).
- `run_heavy_consolidation` now calls `run_weekly_pass`.

Verified end-to-end against an isolated scratch store: a live user capture produces a real `formality_score_weekly` event, and seeding a clear over-formal trend across 5 synthetic weeks followed by a real `run_heavy_consolidation` call detects the trend and relaxes `camouflaging_relaxation` from 0.0 to 0.1, using the project's own unmodified trigger thresholds.

tom-shields-stitch:
Fixes #101.

AUTIST-13 (`camouflaging_relaxation`) has a complete, already-tested detection pipeline in `iai_mcp.camouflaging`: `record_user_formality` scores a turn's formality and logs a `formality_score_weekly` event, and `run_weekly_pass` aggregates that event stream and relaxes the register on a detected over-formal trend. Every pure function here has thorough coverage in `tests/lilli/test_camouflaging_detection.py` — but neither end of the pipeline had a caller anywhere in the package. `capture_turn` never called `record_user_formality`, so the event stream `run_weekly_pass` depends on stayed permanently empty; and nothing called `run_weekly_pass` itself, so even seeded data would never have been aggregated.

Same root-cause shape as #99 and #100: the detection logic was built and tested, the callers were not.

## Fix

- `capture_turn` now calls `record_user_formality` for live, non-replayed user turns (same `live_turn`/`role`/`tier` gate the existing foresight-refresh call right above it already uses, so a historical backfill can't skew weeks that never happened live).
- `run_heavy_consolidation` now calls `run_weekly_pass`. Cadence is nominally weekly, but calling it every heavy pass is harmless — `detect_camouflaging` gates internally on having enough weekly samples and a real trend.

## Verification

Verified end-to-end against an isolated scratch store: a live user capture now produces a real `formality_score_weekly` event (collection side), and seeding a clear over-formal trend across 5 synthetic weeks followed by a real `run_heavy_consolidation` call detects the trend and relaxes `camouflaging_relaxation` from 0.0 to 0.1 (detection + application side) — using the project's own unmodified trigger thresholds, not adjusted for the test.

Also confirmed on a real install: with 33k+ episodic records across 5 days of heavy multi-host usage, all 11 procedural knobs sat at their exact seed defaults (diffed live values against `knobs.py`'s `default_state()` line by line) — consistent with this pipeline never having produced a single signal.

## Test plan

- [x] `pytest tests/test_capture.py tests/test_sleep.py tests/lilli/test_camouflaging_detection.py` — 39 passed
- [x] New `test_live_user_capture_records_formality_signal` (capture side, including a check that a replayed/bulk-imported turn does NOT record a signal)
- [x] New `test_run_heavy_consolidation_relaxes_camouflaging_register` (consolidation side)

### DIFF
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -10,7 +10,7 @@ import shutil
 import threading
 import time
 from collections import deque
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 from uuid import UUID, uuid4
@@ -661,13 +661,26 @@ def capture_turn(
         except Exception as exc:  # noqa: BLE001 -- anticipation is additive
             log.debug("foresight refresh skipped: %s", exc)
 
-        # Feed the camouflaging_relaxation weekly pass (AUTIST-13): scores
-        # this turn's formality and logs a formality_score_weekly event.
-        # Only live, non-replayed user turns count toward the trend, or a
-        # historical backfill would skew weeks that never happened live.
+    # Feed the camouflaging_relaxation weekly pass (AUTIST-13): scores this
+    # turn's formality and logs a formality_score_weekly event.
+    #
+    # Deliberately NOT gated on live_turn like the foresight refresh above:
+    # live_turn=True is only ever passed from the explicit memory_capture
+    # RPC, never from deferred_drain_worker's drain loop -- the path every
+    # host hook actually uses for ambient capture -- so gating on it here
+    # left this permanently dead against real usage; verified zero
+    # formality_score_weekly events existed after days of real multi-host
+    # activity. A recency check on the turn's own timestamp is the signal
+    # that was actually intended: normal ambient capture (drained within
+    # seconds to minutes of the real exchange) is "now" for this purpose; a
+    # historical bulk backfill is not, and must not skew weeks that never
+    # happened live. Mirrors run_light_consolidation's identical 1-hour
+    # recency gate in sleep.py for the same class of concern.
+    if role == "user" and tier == "episodic":
         try:
-            from iai_mcp.camouflaging import record_user_formality
-            record_user_formality(store, text, lang=_formality_lang_signal())
+            if (datetime.now(timezone.utc) - now) < timedelta(hours=1):
+                from iai_mcp.camouflaging import record_user_formality
+                record_user_formality(store, text, lang=_formality_lang_signal())
         except Exception as exc:  # noqa: BLE001 -- signal collection is additive
             log.debug("formality signal recording skipped: %s", exc)
 
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -630,6 +630,16 @@ def capture_turn(
         except Exception as exc:  # noqa: BLE001 -- anticipation is additive
             log.debug("foresight refresh skipped: %s", exc)
 
+        # Feed the camouflaging_relaxation weekly pass (AUTIST-13): scores
+        # this turn's formality and logs a formality_score_weekly event.
+        # Only live, non-replayed user turns count toward the trend, or a
+        # historical backfill would skew weeks that never happened live.
+        try:
+            from iai_mcp.camouflaging import record_user_formality
+            record_user_formality(store, text, lang="en")
+        except Exception as exc:  # noqa: BLE001 -- signal collection is additive
+            log.debug("formality signal recording skipped: %s", exc)
+
     return {"status": "inserted", "record_id": str(rec.id), "reason": f"tier={tier}"}
 
 
--- a/src/iai_mcp/capture.py
+++ b/src/iai_mcp/capture.py
@@ -361,6 +361,37 @@ def _is_episodic_conversational(tier: str, role: str) -> bool:
     return tier == "episodic" and role in {"user", "assistant"}
 
 
+def _formality_lang_signal() -> str:
+    """Best-effort language signal for record_user_formality.
+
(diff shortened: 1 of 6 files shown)


CodeAbra:
Thank you for this, and the 11-knobs-at-seed-defaults check on a real 33k store is the most convincing part of the report. The `live_turn` gate is exactly the right one.

One thing to fix: `lang="en"` is hardcoded, and 3.0.0 made that wrong. `formality_score` returns a neutral 0.5 for a language it has no markers for, which is honest. Passing `"en"` for Russian or Japanese text runs English markers over it, finds nothing, and feeds a fake score into the trend, which can relax the register on no evidence.

In fairness, you copied that hardcoded `en` from our own code, where records were always tagged English. The language pack broke that assumption three days ago, so this is on us to have flagged.

Pass the record's real language, and decide what happens when it isn't supported. Neutral is fine, pretending it's English isn't. Separately, `run_weekly_pass` runs on every heavy pass rather than weekly — probably fine given the internal thresholds, but worth showing that it no-ops.

tom-shields-stitch:
Appreciate the context on where that assumption came from — pushed a new commit fixing both the language claim and something it surfaced while I was verifying the fix.

**Language**: new `_formality_lang_signal()` reads the same config `iai lang status` reads. Empty `languages` (the default) is genuinely English-only, so `"en"` stays correct there. The moment a language pack is enabled, capture still has no per-message language detection, so it now returns a code outside `formality_score`'s marker set and lets the existing neutral-fallback path handle it, rather than guessing.

**The second thing**: while writing a test to confirm calling this repeatedly is a safe no-op, it wasn't — `run_weekly_pass` re-reads the same static event window every time with no memory of what it already processed, so two calls against one unchanged 5-event window detected the same trend twice and relaxed the register twice (`0.0 → 0.1 → 0.2`) on zero new evidence. Same shape as the #104 issue, just one level up. Fixed by tracking the id of the newest `formality_score_weekly` event already processed and skipping re-detection when nothing new has arrived — confirmed a genuinely new sample still triggers further relaxation correctly.

Three new tests: language defaults to `"en"` with no pack, avoids claiming `"en"` once one's enabled, and the dedup (repeated calls are a true no-op, new evidence still works).

tom-shields-stitch:
Found and fixed a real bug in my own previous commit while doing an unrelated end-to-end check of a local install: `record_user_formality` never actually fired for real ambient capture.

I'd gated it on `live_turn`, copying the condition from the foresight-refresh call right above it. Turns out `live_turn=True` is only ever passed from the explicit `memory_capture` RPC — never from `deferred_drain_worker`'s drain loop, which is the path every host hook actually uses. Confirmed on a real install: zero `formality_score_weekly` events existed after several days of genuine multi-host usage since the previous commit deployed.

Pushed a fix: the signal collection now runs off a recency check on the turn's own timestamp instead (mirrors `run_light_consolidation`'s identical 1-hour gate in `sleep.py`), decoupled from `live_turn` entirely. Verified all three real shapes on an isolated store — the ambient drain-worker shape now records, the explicit `memory_capture` shape still records unchanged, and a historical-backfill shape still correctly doesn't record. Also rewrote the test that was previously asserting the buggy behavior as correct (it checked `live_turn=False` + an implicit recent timestamp must not record — that was testing the bug, not guarding against it).

CodeAbra:
Hi, and thank you for turning this around so quickly — the language handling you added was exactly right.

I've decided to take the masking-detection feature out of the engine: the `camouflaging_status` tool, the `camouflaging_relaxation` knob, and the formality-relaxation it applied on the recall path. So I'm going to close this one — and to be clear, it's the feature being removed, not anything about your code.

The reason: after testing the knob, it didn't prove its worth. The register adjustment it produced didn't make recall measurably better. And beyond that, this knob quietly reshapes the assistant's tone based on how the user writes, and that's the opposite of what this project is meant to be. The memory is supposed to observe and preserve — direct, unmasked, a stable identity that doesn't drift to match the person in front of it. A feature that adjusts the register to blend in works against that, so it doesn't belong in the product, however well it's built.

Separately, the dead-code paths you flagged — `should_bootstrap_trigger`, `session_exit`, and a couple more that surfaced afterward — are being fixed right now. That was a genuinely useful catch, and I appreciate the time you put into it. You'll be credited in the changelog for it, and I'll be adding you to the project's contributors.

PR #108, opened 2026-08-16
Windows support: fix daemon boot (SIGHUP), IPC listener, and ambient hooks; plus reliability fixes

Kmmille2:
## Summary

The daemon doesn't currently start on Windows — two Python-portability bugs kill it before it can serve, so the store stays empty and MCP hosts can't reach it. This PR fixes those, makes the ambient-capture hooks work under Git Bash, and includes three reliability fixes I hit while getting it running.

Verified end-to-end on **Windows 11 / Python 3.11** (from-source build, MSVC toolchain): the daemon boots, binds its IPC listener, and `capture → recall` returns records through the MCP path in both Claude Code and Codex. Backfilled ~480 records and confirmed recall + the `iai brain` dashboard.

## Windows portability

- **Daemon boot crash — `signal.SIGHUP` is absent on Windows.** `_install_boot_signal_trace` built the tuple `(SIGTERM, SIGINT, SIGHUP)`; evaluating `signal.SIGHUP` raised `AttributeError` *before* the per-signal `try/except` could catch it, killing boot. Resolve each signal lazily via `getattr` — mirrors the run-loop shutdown handler's existing guard.
- **IPC listener never binds — `asyncio.start_unix_server` is absent on Windows.** `SocketServer.serve()` read `inspect.signature(asyncio.start_unix_server)` *before* the `IS_WINDOWS` branch; on Windows that attribute doesn't exist, so `serve()` raised immediately and — as a `create_task`'d coroutine — the exception was never retrieved. The daemon ran but was unreachable (no `.daemon.port`; `daemon status` reported not running). Moved the POSIX-only signature read past the Windows early return.
- **Ambient-capture hooks hardcoded `/usr/bin/python3`**, which doesn't exist under Git Bash, so capture silently no-opped (fail-safe `exit 0`). Resolve the interpreter portably: honor `IAI_MCP_PYTHON`, skip the Windows Store `python3` App-Execution-Alias stub, fall back to the `py` launcher.
- **Pinned `*.sh` to LF** via `.gitattributes` so the hooks run under bash on Windows (autocrlf otherwise materializes CRLF → bad interpreter).
- **Console `CTRL_C` survival.** A Task-Scheduler daemon attached to the interactive console was terminated with `STATUS_CONTROL_C_EXIT` when the host CLI restarted. Install a Windows console-ctrl handler (only under `IAI_MCP_LAUNCHD_MANAGED`) that swallows CTRL_C/CTRL_BREAK; Task Scheduler still stops it via `TerminateProcess`.

## Reliability (general, not Windows-specific)

- **Poison-pill capture drain.** A record `capture_turn` rejects (e.g. an embedder failure) raised out of both drain loops, so the spool file never unlinked and every subsequent drain re-hit the same record — silently wedging all future capture. Quarantine-skip the failing record so the file still unlinks and the backlog drains (re-runs stay loss-free via the `source_uuid` idem-tag).
- **RO-pool slot leak in `recycle()`.** The `except queue.Full` branch closed a slot without decrementing `_filled` (unlike `_release`); a race with a concurrent `_release()` leaks a slot, and over a daemon lifetime the pool drains until `borrow()`'s `get()` blocks forever. Decrement under `_fill_lock`, matching `_release()`.
- **Corrupt-page host abort/panic in `lillibrain`.** `read_overflow_chain` reserved `Vec::with_capacity(total_len)` from a caller-supplied on-disk length — a corrupted/forged varint hits the allocator abort path (uncatchable, kills the host). `cursor.rs` `payload_of`/`read_full_cell` sliced `page[start..start+len]` raw from the same unvalidated decode (panics across PyO3). Cap the reservation at the physical maximum; use `checked_add` + `.get().ok_or(Integrity)` so a damaged page yields a typed error, as `raw_leaf_cell_bytes` already does.

Happy to split this into two PRs (Windows portability vs reliability) if you'd prefer smaller reviews. Thanks for the project — it's a genuinely nice piece of engineering.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+
+# iai-pme (windows-adapt): shell hooks MUST stay LF or bash fails on Windows
+*.sh text eol=lf
--- a/crates/lillibrain/src/btree/cursor.rs
+++ b/crates/lillibrain/src/btree/cursor.rs
@@ -156,7 +156,23 @@ impl<'p> Cursor<'p> {
             let first_ovf = read_spilled_pointer(page, raw.payload_start)?;
             read_overflow_chain(self.pager, first_ovf, raw.payload_len)
         } else {
-            Ok(page[raw.payload_start..raw.payload_start + raw.payload_len].to_vec())
+            let end = raw
+                .payload_start
+                .checked_add(raw.payload_len)
+                .ok_or_else(|| StoreError::Integrity {
+                    detail: "inline cell payload length overflow".to_string(),
+                })?;
+            Ok(page
+                .get(raw.payload_start..end)
+                .ok_or_else(|| StoreError::Integrity {
+                    detail: format!(
+                        "inline cell payload {}..{} out of page bounds (page len {})",
+                        raw.payload_start,
+                        end,
+                        page.len()
+                    ),
+                })?
+                .to_vec())
         }
     }
 
@@ -600,10 +616,24 @@ pub fn read_full_cell(pager: &Pager, page: &[u8], idx: usize) -> Result<(i64, Ve
         let payload = read_overflow_chain(pager, first_ovf, raw.payload_len)?;
         Ok((raw.key, payload))
     } else {
-        Ok((
-            raw.key,
-            page[raw.payload_start..raw.payload_start + raw.payload_len].to_vec(),
-        ))
+        let end = raw
+            .payload_start
+            .checked_add(raw.payload_len)
+            .ok_or_else(|| StoreError::Integrity {
+                detail: "inline cell payload length overflow".to_string(),
+            })?;
+        let payload = page
+            .get(raw.payload_start..end)
+            .ok_or_else(|| StoreError::Integrity {
+                detail: format!(
+                    "inline cell payload {}..{} out of page bounds (page len {})",
+                    raw.payload_start,
+                    end,
+                    page.len()
+                ),
+            })?
+            .to_vec();
+        Ok((raw.key, payload))
     }
 }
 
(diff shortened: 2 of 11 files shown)


CodeAbra:
Really nice work, Kyle, and it cleared our security review. We'll take the Windows fixes and merge them under your name.

The three engine hardening commits are good too. They'll come in through our storage-core pipeline rather than this PR, since we keep a single writer on that code, with your authorship on them.

The drain poison-pill guard is the one we can't take as-is: skip-and-unlink turns a stuck backlog into silent data loss, and a systemic failure like a cold embedder or a full disk would quietly wipe the whole backlog. If you want to rework it, quarantine the failing event to a `.permanent-failed-*.jsonl` sidecar, count it under a new `failed` key instead of `skipped`, and stop the drain past a small failure threshold. We already have a recovery command and a doctor check for those, and I'd be glad to review it.

Two small asks so I can merge the Windows part: rebase on latest main (3.0.2 just shipped and fixed the snapshot path your commits touch), and drop the bench-CSV churn the `.gitattributes` pulled in. Keep the `.gitattributes` itself.

CodeAbra:
Really appreciate how thorough this is — you got the daemon booting on Windows end-to-end. Two things now that 3.0.5 landed: (1) it reworked the daemon internals, so this now conflicts and needs a rebase; (2) it overlaps #126, which is more surgical and ships tests for the same boot path, so I'm planning to base the Windows daemon-boot fix on that. Would you be up for narrowing this to the unique reliability fixes #126 doesn't cover? Separately, the Rust engine changes (`crates/lillibrain/src/btree/…`) should go through a review from our engine side rather than riding in with the Windows fixes — happy to coordinate. Credit for the boot diagnosis stands regardless.

CodeAbra:
Thank you so much! This is really valuable. Can you please rebase so I can merge?

CodeAbra:
Hey Kyle — any luck with the rebase? The branch still shows conflicts against main, so I can't merge it yet. Happy to help if you're stuck. Once it's rebased on top of current main I'll get it in.

PR #111, opened 2026-08-17
fix(install): honor detected Python instead of bare python3 for venv

### LINKED ISSUE #110: install.sh creates the venv with bare python3, ignoring the interpreter bootstrap.sh detected
## What happens

`scripts/bootstrap.sh` preflight probes for a supported interpreter (`python3.12` → `python3.11` → `python3`, gated on 3.11/3.12) and reports success, e.g.:

```
✓ Python 3.11.16 at /opt/homebrew/bin/python3.11
```

But it never passes that interpreter to `scripts/install.sh`, which creates the venv with a bare `python3 -m venv .venv`. On a system where `python3` resolves to 3.13+ (Homebrew currently ships 3.14 as `python3`), the install fails after the Rust prerequisite check with:

```
==> editable install (pip -e .)
ERROR: Package 'iai-pme' requires a different Python: 3.14.7 not in '<3.13,>=3.11'
```

## Reproduce

On macOS with Homebrew Python 3.14 as default `python3` and `python3.11` also installed:

```
bash scripts/bootstrap.sh
```

Preflight passes (it found 3.11), venv is built on 3.14, `pip install -e .` dies.

There's a second trap: because `install.sh` reuses an existing `.venv` without checking its interpreter, re-running after installing a correct Python still fails until the user knows to delete `.venv` manually.

## Suggested fix

Make `install.sh` own interpreter selection (env override + the same version-gated probe bootstrap uses), have `bootstrap.sh` pass its detected interpreter down, and reject an existing `.venv` built on an unsupported Python with a clear message. PR incoming.

anubissbe:
Fixes #110.

## Problem

`bootstrap.sh` preflight detects a supported interpreter (3.11/3.12), but `install.sh` builds the venv with whatever bare `python3` resolves to. With Homebrew's `python3` now at 3.14, the install fails at `pip install -e .` with `requires a different Python: 3.14.7 not in '<3.13,>=3.11'` — after the multi-minute Rust prerequisite step. And since `install.sh` reuses an existing `.venv` unchecked, re-runs keep failing until the user deletes `.venv` by hand.

## Changes

- `install.sh` now selects the interpreter itself: `IAI_PYTHON` env override first (validated), otherwise the same version-gated `python3.12` → `python3.11` → `python3` probe bootstrap uses, with a clear `die` if none qualifies.
- An existing `.venv` on an unsupported interpreter is rejected up front with instructions to remove it, instead of failing later inside pip.
- `bootstrap.sh` passes its detected interpreter down via `IAI_PYTHON=` (dry-run output updated to match).

`install.sh` stays independently runnable for collaborators (probe fallback), and the fix is a no-op on systems where `python3` is already 3.11/3.12.

## Testing

- `bash -n` on both scripts.
- `bootstrap.sh --dry-run` shows `would run: IAI_PYTHON=/opt/homebrew/bin/python3.11 bash .../install.sh`.
- Negative test: `IAI_PYTHON=$(command -v python3)` (3.14.7) → dies immediately with `IAI_PYTHON=/opt/homebrew/bin/python3 is Python 3.14.7; need Python 3.11 or 3.12`.
- Real-world repro of the underlying bug: full bootstrap on macOS arm64 with Homebrew python3 = 3.14.7 failed exactly as in #110; with the venv on 3.11.16 the same install completes and `iai-mcp doctor` passes all checks.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -126,10 +126,10 @@ fi
 # ---------------------------------------------------------------------------
 step "building (a few minutes: the Rust engine compiles from source)"
 if [ "${DRY_RUN}" = "1" ]; then
-    printf '   \033[0;36m→\033[0m would run: bash %s/scripts/install.sh\n' "${SRC_DIR}"
+    printf '   \033[0;36m→\033[0m would run: IAI_PYTHON=%s bash %s/scripts/install.sh\n' "${PYTHON}" "${SRC_DIR}"
 else
     [ -f "${SRC_DIR}/scripts/install.sh" ] || die "scripts/install.sh missing from the clone"
-    bash "${SRC_DIR}/scripts/install.sh"
+    IAI_PYTHON="${PYTHON}" bash "${SRC_DIR}/scripts/install.sh"
 fi
 
 IAI_BIN="${SRC_DIR}/.venv/bin/iai-mcp"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -40,11 +40,29 @@ else
     # 1. venv
     # -----------------------------------------------------------------------
     step "python venv"
-    if [ ! -d .venv ]; then
-        python3 -m venv .venv
-        ok ".venv created"
+    # The package supports 3.11 and 3.12 only. IAI_PYTHON overrides detection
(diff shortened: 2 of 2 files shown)


CodeAbra:
Github is down. Will merge after this solved. Thank you so much for the contribution 🙏🏼

Issue #119, opened 2026-08-18
cowork install/status: home-discovery heuristic doesn't match current Desktop app storage, confirmed empirically after a full relaunch


oscampo:
## Summary

`iai-mcp cowork install`/`cowork status` never find a Cowork home on a real, actively-used Windows install, even after a full quit, relaunch, and brand-new Cowork session, exactly the remedy `cmd_cowork_install` itself suggests ("rerun install after the first Cowork session"). This is a different bug than #114: #114 is about the combined `capture-hooks status` output being misleading about what "WIRED" means. This one is about `_discover_cowork_homes()` looking for files that this Desktop app version apparently never writes, regardless of usage history.

Follows up on #113 and #114, same machine, same install.

Environment: Windows 11, Claude Desktop `appVersion 1.32352.0` (electron 42.9.2), `iai_pme 3.0.1`.

## What the code checks

From `src/iai_mcp/cli/_cowork.py`:

```python
def _cowork_session_roots() -> list[Path]:
    ...
    if os.name == "nt":
        appdata = os.environ.get("APPDATA")
        base = Path(appdata) if appdata else home / "AppData" / "Roaming"
        return [base / "Claude" / "local-agent-mode-sessions"]
    ...

def _looks_like_cowork_home(d: Path) -> bool:
    return (
        (d / "cowork_settings.json").exists()
        or (d / "cowork_plugins").is_dir()
        or (d / ".claude.json").exists()
    )
```

It walks `local-agent-mode-sessions/<account-uuid>/<device-uuid>/` and only counts a directory as a home if one of those three markers exists directly inside it.

## What's actually on disk

Only one account uuid exists under `local-agent-mode-sessions` (`fc1855e9-...`), with one device uuid inside it (`e15b7a21-...`). That directory is demonstrably the live, actively-used home, not a stale one: it contains today's session data, `rpm/` (the plugin cache) has entries like `rpm/plugin_01J1ZUJcofzWJxiajSBsZ8US/.claude-plugin/plugin.json` matching every marketplace plugin currently enabled in the session (cockroachdb, zoom-plugin, etc., visible in the running Cowork skills list).

`Test-Path` for the three files `_looks_like_cowork_home` checks for, run directly against that directory: `cowork_settings.json` → `False`, `cowork_plugins` → `False`, `.claude.json` → `False`.

[...]

PR #120, opened 2026-08-18
Fix #113 (README doc gaps) and #114 (misleading capture-hooks status)

### LINKED ISSUE #113: Windows: 3 boot-blocking bugs (fixed, commits linked) + 2 README/doctor doc gaps found verifying the README's own claims
## Summary

Three real Windows boot-blocking bugs, found and fixed with a traceback and `py-spy` in hand (not guessed at), plus two README/code documentation gaps found while independently verifying the README's own claims against a live Windows 11 install. All three code bugs are already fixed and merged on a fork (`oscampo/iai-personal-memory-engine`), with commit links below for anyone hitting the same wall. Opening this mainly so the fixes are visible upstream and the doc gaps get tracked, happy to open a PR against `main` if that's welcome.

Environment: Windows 11, Python 3.12, `iai-pme` installed via `pip install .` from a git checkout (not the PyPI wheel).

## Bug 1 — daemon crashes at boot: `AttributeError: module 'signal' has no attribute 'SIGHUP'`

`src/iai_mcp/daemon/__init__.py`, `_install_boot_signal_trace()`: the tuple `(signal.SIGTERM, signal.SIGINT, signal.SIGHUP)` was built directly in the `for` clause, evaluated *before* the inner `try/except` could guard it. `SIGHUP` doesn't exist on Windows, so this raised immediately and killed the boot, before the graceful-shutdown handlers a few lines below (which already correctly used `getattr(signal, "SIGHUP", None)`) ever installed.

Fix: build the signal list with the same `getattr` guard already used elsewhere in the same file.

Commit: `oscampo/iai-personal-memory-engine@7e576682`

## Bug 2 — per-turn capture hook fails on every turn on Windows (`rc=127`)

Both copies of the capture hook (`plugin/hooks/iai-mcp-turn-capture.sh` and `src/iai_mcp/_deploy/hooks/iai-mcp-turn-capture.sh`, which must stay byte-identical per the existing "Sync the plugin hook copy" commit convention) hardcoded `/usr/bin/python3`, which doesn't exist under Git Bash on Windows, and unconditionally `import fcntl`, which is POSIX-only.

Fix: resolve `IAI_MCP_PYTHON` / `python3` / `python` on `PATH` instead of a fixed path (fail-safe: no interpreter found → exit 0, same as any other capture failure), and replace the unconditional `fcntl` import with a small self-contained `msvcrt.locking`-based shim inside the embedded `PY_SCRIPT` (mirrors `iai_mcp._flock`, inlined because this script runs under whatever `python3` resolves on `PATH`, not necessarily the venv where the package is installed).

Commits: `oscampo/iai-personal-memory-engine@c94e996f`, `oscampo/iai-personal-memory-engine@b31a489b`

## Bug 3 — daemon boots but the socket never opens, no traceback anywhere

The real one behind the "atascado sin traceback" symptom. `SocketServer.serve()` (`src/iai_mcp/socket_server.py`) called `inspect.signature(asyncio.start_unix_server)` unconditionally, *before* the `IS_WINDOWS` branch further down in the same function. `asyncio.start_unix_server` doesn't exist on Windows, so this raised `AttributeError` immediately — and because `serve()` runs as an un-awaited `asyncio.create_task()` in `daemon/__init__.py`'s `main()`, the exception died completely silently. The rest of the daemon (store, watchdog, queues) kept running, `daemon status` just said "not running", `.daemon.port`/`.daemon.token` never got created, and nothing anywhere logged a traceback. Confirmed with `py-spy` on a live stuck process: `MainThread` was idle in `run_forever`, not stuck mid-boot — the task had already died.

Fix: guard the `inspect.signature` call, plus a `done_callback` on `mcp_socket_task` that now logs any future exception at ERROR level with the full traceback and triggers normal shutdown, so a boot-time failure in the socket server can't hide behind an otherwise-healthy-looking daemon again.

Verified end to end on the real target machine after the fix: `daemon status` → `ok: True`, `state: WAKE`, `.daemon.port`/`.daemon.token` created; `iai capture` + `iai recall` round-tripped `via daemon`, not the store-direct fallback.

Commits: `oscampo/iai-personal-memory-engine@1107640e`, `oscampo/iai-personal-memory-engine@23727ff0`

## Doc gap 1 — "Deployment surface" claims Unix-socket-only, false on Windows

> MCP transport is a Unix domain socket. No TCP listener, no bind address, no auth surface to misconfigure.

`src/iai_mcp/_ipc.py` implements a TCP loopback listener (`127.0.0.1:<ephemeral port>`) with token auth via `icacls` specifically for Windows (already merged, predates 3.0.0). Confirmed live: `.daemon.port` and `.daemon.token` both exist and are in active use on a working Windows install. The "Windows: beta" badge at the top of the README doesn't cover this — the "Deployment surface" section states the Unix-socket claim as a flat, platform-unqualified fact. Loopback + token is a reasonable design, it's just not what that sentence says is happening.

## Doc gap 2 — `doctor` table says 27 checks, the actual command runs 30

The "All 27 checks, one line each" table (under `## Doctor`) lists exactly 27 rows (`a`–`z` plus `+`). A live `iai-mcp doctor` run on the same install shows three additional checks that don't appear in the table at all:

- `(ii) store embed identity`
- `(aa) capture-state hygiene`
- `(bb) nightly insight mint`

Not a stale count, three real checks are just missing from the docs entirely.

---

Happy to send a PR for the doc fixes (and rebase the three Windows commits from the fork onto a clean branch against `main`) if that's the preferred path — let me know.

### LINKED ISSUE #114: capture-hooks status: "Claude Desktop: WIRED" is misleading, only Cowork has ambient hooks, Chat tab has none structurally
## Summary

`iai-mcp capture-hooks status` reports `Claude Desktop: ... WIRED` and folds that into the overall `status: ACTIVE` line, which reads as "ambient capture works in Claude Desktop." It doesn't, for the surface most people mean when they say "Claude Desktop": the plain Chat tab. Verified live against a real install, not from reading the source.

## What "WIRED" actually checks

`capture-hooks status` for the Desktop line only checks that `iai-mcp` is registered under `mcpServers` in `claude_desktop_config.json`. Confirmed by reading that file directly on a working install: it contains `mcpServers` (`github`, `iai-mcp`) and nothing else, no `hooks` key, no per-turn event config of any kind. That's correct and sufficient for MCP *tools* (`memory_capture`, `episodes_recent`, etc.) to be available for a model to call on request. It says nothing about whether anything captures automatically.

## What was actually tested

Ran a full working session in Claude Desktop's **Chat tab** (`capture-hooks status` showing `WIRED` / `ACTIVE` the whole time), covering roughly two hours of substantive back-and-forth. Queried `episodes_recent` afterward: **zero** records from that conversation existed in the store. Only records from earlier, unrelated manual `iai capture` CLI calls were present. Chat tab has no hook surface to capture from, because Claude Desktop's config format (unlike Claude Code's `settings.json`) doesn't expose `UserPromptSubmit`/`Stop`-style events to local scripts at all — this isn't a missing wire, it's a surface that doesn't exist for that tab.

## The nuance that makes this worth a precise fix, not just a caveat

**Cowork** (a different mode inside Claude Desktop) *does* have a real, separate hook mechanism, via the already-existing `iai-mcp cowork install/status` subcommand, which wires into each Cowork home's `cowork_settings.json`. Confirmed by running `cowork install` live: it correctly staged the plugin and reported `no Cowork homes found` because Cowork had never been opened on that machine, exactly the failure mode you'd want it to report. Once a Cowork session exists and install is rerun, that surface should genuinely get ambient capture, same reliability class as Claude Code's hooks.

So the current single `Claude Desktop: ... WIRED` line conflates three different things under one label:
1. MCP tools available in Chat tab (true, and useful, but not ambient capture)
2. MCP tools available in Cowork (same as above)
3. Ambient hook capture in Cowork specifically (separate subcommand, separate state, can be INACTIVE even when Desktop's `WIRED` line reads true)

## Suggested fix

Split the Desktop line in `capture-hooks status` output into something like:
```
Claude Desktop MCP (tools, Chat + Cowork): WIRED
Claude Desktop Cowork hooks (ambient capture): ACTIVE / INACTIVE  (see: iai-mcp cowork status)
```
and drop "Desktop also wired" from the summary `status: ACTIVE` line unless Cowork hooks are specifically active, since that's the only Desktop surface where "wired" implies the same thing it implies for Claude Code.

Happy to send a PR for the status-output wording if that's welcome, this doesn't require any behavior change, only making the existing INACTIVE/ACTIVE distinction (which the tool already computes correctly for `cowork status`) visible from the single combined `capture-hooks status` command too.

oscampo:
## Qué arregla

Closes #113 (the doc portion — the three Windows boot bugs from that issue were already fixed upstream via #111, this closes the remaining README gaps) and closes #114.

1. **README**: the `doctor` table documented 27 checks, the command runs 30. Added `(ii) store embed identity`, `(aa) capture-state hygiene`, `(bb) nightly insight mint` in their actual `run_diagnosis()` order, updated the count in all three places that cited "27 checks". The "Deployment surface" section stated the transport is always a Unix domain socket with no exception; it now documents that Windows uses TCP loopback with a per-boot random auth token (`_ipc.py`) instead of contradicting the daemon's own behavior there.
2. **`capture-hooks status`**: the single `Claude Desktop: ... WIRED` line conflated "MCP registered in `claude_desktop_config.json`" with "ambient capture active" — the Chat tab has the former but never the latter (confirmed live: a long real session produced zero new captures, since Chat has no hook mechanism at all). Now prints two separate lines, with the capture-status one delegating to the same Cowork-home discovery `cowork status` already uses, so it reflects real hook wiring rather than just MCP-entry presence. The `status: ACTIVE` summary no longer claims "Desktop also wired" unless Cowork ambient capture is actually active.

## Verification

Ran on the real target Windows machine, not just reviewed in diff:

- `iai-mcp doctor`: 30 real checks, `(ii)`/`(aa)`/`(bb)` present and functioning.
- `iai-mcp capture-hooks status`: both new lines print correctly; since that machine currently has no discoverable Cowork home (a separate, Desktop-app-side issue, not something this project can fix — see #119 for the full investigation, including a control test against this exact repo), it correctly reports "no Cowork homes found — skipped" and the summary no longer overclaims about Desktop.
- `pytest tests/test_capture_hooks_install_registers_turn_hook.py tests/test_capture_hooks_install_registers_per_turn_recall.py tests/test_cowork_install.py`: 10 passed, 5 skipped, 0 failed.

### DIFF
--- a/HANDOFF.md
+++ b/HANDOFF.md
@@ -1,0 +1,111 @@
+# Handoff → Code
+
+## Origen
+
+Conversación en Chat, sesión del 17 de agosto de 2026. Empezó revisando un
+repo de terceros (`valera-studio-group/valera-studio-project`, un harness de
+memoria paginada para Claude Code CLI) como posible alternativa mientras se
+resolvía un problema real: iai-mcp no arrancaba en Windows.
+
+## Contexto
+
+Ayer (sesión previa, no en este hilo) se diagnosticó que el daemon de
+iai-mcp crasheaba al bootear en Windows 11 con un traceback de
+`AttributeError: module 'signal' has no attribute 'SIGHUP'`. Tras aplicar un
+parche local, el daemon llegó más lejos pero se quedó atascado (CPU casi
+plana, sin nuevo traceback, sin socket abierto) durante varios minutos, sin
+llegar a confirmarse healthy. Esa sesión se detuvo ahí, sin escalar más a
+ciegas.
+
+En este hilo se comparó el fork (`oscampo/iai-personal-memory-engine`)
+contra el proyecto original (`CodeAbra/iai-personal-memory-engine`) y se
+encontraron y corrigieron dos bugs reales, ya subidos a `main` del fork.
+
+## Hallazgos verificados
+
+- [Seguro] El fork de Oscar no tenía ningún commit propio: estaba al día
+  con `upstream/main` pero además 5 commits *atrás* (Release 3.0.2 y
+  siguientes). El parche de `SIGHUP` de la sesión de ayer nunca llegó a
+  commitearse, vivía solo (si acaso) en el working tree local de la máquina
+  Windows.
+- [Seguro] El bug de `SIGHUP` existe también en `upstream/main`, sin
+  arreglar: no es un problema del fork ni del entorno de Oscar, es un bug
+  real y reportable en `CodeAbra/iai-personal-memory-engine`.
+- [Seguro] El bug real: en `src/iai_mcp/daemon/__init__.py`, la función
+  `_install_boot_signal_trace` construía la tupla
+  `(signal.SIGTERM, signal.SIGINT, signal.SIGHUP)` directamente en la
+  cláusula `for`, evaluada antes de que el `try/except` interno pudiera
+  protegerla. Eso crashea en Windows (sin `SIGHUP`) antes de que los
+  handlers de shutdown graceful, unas líneas más abajo, lleguen a
+  instalarse (esos sí ya usaban `getattr(signal, "SIGHUP", None)`).
+- [Seguro] Segundo bug, idéntico en dos copias que deben mantenerse
+  sincronizadas (`plugin/hooks/iai-mcp-turn-capture.sh` y
+  `src/iai_mcp/_deploy/hooks/iai-mcp-turn-capture.sh`): el hook de captura
+  por turno invocaba `/usr/bin/python3` con ruta fija (no existe en Git
+  Bash de Windows) e importaba `fcntl` sin guard (módulo POSIX-only). Esto
+  causaba `rc=127` en cada turno en Windows, es decir, la captura ambiental
+  ("nunca digo 'recuérdalo'") estaba rota de fondo en esa plataforma.
+- [Seguro] El transporte del daemon para Windows **sí existe** en el
+  código: `src/iai_mcp/_ipc.py` implementa TCP loopback
+  (`127.0.0.1:<puerto efímero>`) con autenticación por token vía `icacls`,
+  agregado en el commit `6bfc5d9` ("Add Windows support..."), ya en `main`
+  desde antes del Release 3.0.0. Esto contradice la lectura inicial de que
+  "el soporte Windows no existía"; lo que se vio ayer (atascado sin socket)
+  probablemente era el daemon quedándose en otro punto antes de llegar a
+  bindear esa capa, no una ausencia de la capa misma.
+- [Probable] El atascamiento visto ayer (CPU plana, sin traceback) podría
+  desaparecer solo con el fix de `SIGHUP` si el daemon simplemente no había
+  llegado tan lejos antes. No confirmado: no se ha vuelto a correr en la
+  máquina Windows real desde que se subieron los parches.
+
+## Commits aplicados (ya en `main` de `oscampo/iai-personal-memory-engine`)
+
+1. `7e576682` — guard de `SIGHUP` en `_install_boot_signal_trace`
+   (`src/iai_mcp/daemon/__init__.py`).
+2. `c94e996f` — resolución de intérprete Python + guard de `fcntl` en
+   `plugin/hooks/iai-mcp-turn-capture.sh`.
+3. `b31a489b` — mismo fix, copia sincronizada en
+   `src/iai_mcp/_deploy/hooks/iai-mcp-turn-capture.sh`.
+
+## Conclusión
+
+El camino correcto sigue siendo terminar el port a Windows sobre el fork
+propio, no adoptar una herramienta externa (Valera Studio Harness, un
+wrapper de Claude Code CLI sin componente MCP, evaluado y descartado en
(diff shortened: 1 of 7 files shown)


CodeAbra:
Thanks for this, and for verifying it on a real Windows box. The #114 status fix is exactly right — splitting the one WIRED line into "MCP registered" vs the actual Cowork ambient capture is the correct call, and the comment explaining why is great.

Could you trim the PR down to just that `_capture.py` change though? A few reasons:

- the README doc gaps already got fixed on main a couple days ago. one thing: the real check count is 31, not 30 — the table's still missing `(cc) background liveness` alongside the ii/aa/bb you added. so main's already ahead here.
- `HANDOFF.md` looks like your own local handoff note (the fork PR ref, the branch name) rather than something for the repo, so probably drop that one.
- the daemon/socket_server/hook changes are the #113 boot fixes, and those overlap another open Windows PR (#108). i'm sorting out who owns the Windows engine side so those don't collide, so hold them out of this PR for now and i'll come back to them.

Narrow it to the status fix and i'll merge it under your name. Really appreciate how carefully you've been working through these.

Issue #158, opened 2026-08-31
[bug] `iter_record_columns` raises `disk I/O error` inside the daemon, blocking SCHEMA_MINE and the hippea cascade


farislmn:
**Version:** iai-pme 3.0.8 (pip, PyPI), `iai` CLI 3.0.8
**Platform:** Ubuntu LXC on Proxmox, Linux 6.14.11-9-bpo12-pve, Python 3.12, x86_64 (AVX2 available)
**Store:** stdlib sqlite3 driver, `schema_version=1`, 2218 records, 11.8 MB
**First observed:** 2026-08-17 05:19:55 UTC — the first day in the journal, i.e. since installation
**Frequency:** ~2,800 occurrences per day, every day, unbroken

---

## Summary

Every call into `HippoStore.iter_record_columns()` fails inside the daemon process with
`sqlite3.OperationalError: disk I/O error`, raised on the **first page** of the keyset scan
(`last_id == ""`). Because that method is the single seam for full-corpus column scans, both
callers that depend on it die:

- `lilli/cycle/sleep_pipeline/_schema_mine.py:19` → `induce_schemas_tier0()` → the sleep pipeline
  aborts at `SleepStep.SCHEMA_MINE`, so **no insight is ever minted**;
- `daemon/_watchdog.py:186` → `retrieve.build_runtime_graph()` → the hippea cascade loop fails on
  every iteration.

`iai-mcp doctor` surfaces this only indirectly, as
`[FAIL] (bb) nightly insight mint — 3 consecutive nights without a minted insight`. The underlying
exception is not reported by any check, so on this machine the sleep pipeline appears to have never
completed since install: `.daemon-state.json` still carries
`sleep_cycle_baseline_at = "2026-08-17T04:49:50.775855+00:00"`.

## The key detail

**The failure is specific to the daemon process.** The identical scan succeeds from any other
process on the same machine, against the same live store, while the daemon is running:

- 40 consecutive full keyset scans (`SELECT id, centrality FROM records WHERE id > ? ORDER BY id
  LIMIT 256`, paged to exhaustion, 2218 rows each) from a separate process: **40/40 clean**.
- The same scan driven through the package's own connection seam
  (`iai_mcp.hippo._raw_open.open_store_conn(path, read_only=True)`, which returns `None` here and
  falls through to `_sqlite_stdlib.connect`): **clean, 2218 rows**.
- `PRAGMA quick_check` on `brain.sqlite3`: `ok`.

[...]

PR #160, opened 2026-09-02
Drain idle socket handlers so daemon shutdown cannot hang

farislmn:
Fixes the shutdown hang reported in #158 (finding 3 there).

With any MCP client connected and idle, `systemctl --user stop iai-mcp-daemon` takes the full `TimeoutStopSec` and the daemon dies by SIGKILL. Every time, never cleanly. A SIGKILLed daemon cannot checkpoint or close its WAL, so this may also bear on the other findings in that issue, though I would not assume it.

## Cause

`Server.wait_closed()` returns only once every active connection has been dropped. From the CPython 3.12.3 stdlib source:

> Wait until server is closed and all connections are dropped. If the server is not closed, wait. If it is closed, but there are still active connections, wait.
>
> Historical note: In 3.11 and before, this was broken, returning immediately if the server was already closed, even if there were still active connections.

A `SocketServer.handle` task parked on `await reader.readline()` for an idle client never drops on its own, and nothing tracked or cancelled those tasks. On 3.11 and earlier that did not matter, because `wait_closed()` returned immediately. On 3.12.1+ it became load-bearing, which is why this can appear without any code change on your side. Worth checking against your CI matrix: a suite that runs on 3.11 would not catch it.

## Fix

Track the per-connection handler tasks on the `SocketServer` and cancel them before `wait_closed()`.

The `async with server:` had to go, and this is the part I would most like a second opinion on. `daemon.main()` sets `shutdown_event` and then immediately cancels the serve task (`daemon/__init__.py:3265` then `:3284`). The cancellation lands while `serve` is still parked on `shutdown_event.wait()`, which skips the `with` body entirely and runs `__aexit__` — calling `close()` and `wait_closed()` with the handlers still live. Draining in the body looks correct and still hangs in production. An explicit `try/finally` runs the same teardown on the path cancellation actually takes.

I kept the change to the socket server rather than bounding the `gather` in `daemon.main()` with a timeout, since that would cap the damage without addressing why the handlers never finish. Happy to reshape this if the intended connection lifecycle is different from what I assumed.

## Evidence

Measured on Ubuntu (LXC on Proxmox), Python 3.12.3, one idle client attached for the whole run, same daemon and store each time:

| Build | Idle client | Stop time | Outcome |
| --- | --- | --- | --- |
| stock 3.0.8 | attached | 60s | SIGKILL |
| drain inside the `with` body | attached | 60s | SIGKILL |
| drain in a `finally` (this PR) | attached | 0s | clean |

With no client attached, stock already stops cleanly in about 2s, which is what made this look intermittent at first.

The asyncio task dump during the hang, showing all four live tasks:

```
===== task dump 2026-09-02T08:21:44 shutdown_set=True =====
--- Task-1 done=False ---   coro: 'main'
  daemon/__init__.py:3285   await asyncio.gather(*_cancel_targets, return_exceptions=True)
--- Task-7 done=False ---   coro: 'SocketServer.serve'
  socket_server.py:314      async with server:
--- Task-147 done=False --- coro: 'SocketServer.handle'
  socket_server.py:90       line = await reader.readline()
--- Task-2299 done=False -- coro: 'SocketServer.handle'
  socket_server.py:90       line = await reader.readline()
```

`shutdown_set=True`, so the signal handler and `shutdown.wait()` both work correctly; the teardown is running and stuck five lines in. Two dumps 30s apart are identical, so it is wedged rather than slow. Note `serve` sits on line 314, the `async with`, not on the `wait_closed()` below it — that is what showed the body-drain version could not work.

## Tests

Added `tests/test_socket_shutdown_drains_idle_handlers.py`. It fails on stock (10s timeout, no completion) and passes with the fix.

The test cancels the serve task rather than only setting `shutdown_event`, deliberately: a test that only sets the event passes against the broken body-drain version. That distinction is in a comment so a future refactor does not quietly weaken it.

Existing socket suite on this branch: `tests/test_socket_server_dispatch.py`, `test_socket_concurrent_clients.py`, `test_socket_disconnect_reconnect.py`, `test_socket_activity_excludes_health_probe.py` all pass. `test_socket_fail_loud.py` has 2 failures ("daemon never bound socket within 30s") that reproduce identically on unpatched `main` in my environment, so they look environmental here rather than related to this change.

## Reproducing

Start the daemon, connect a client and send nothing, then stop it:

```python
import socket, os, time
s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s.connect(os.path.expanduser("~/.iai-mcp/.daemon.sock"))
time.sleep(600)
```

Stock hangs for the full stop timeout; with this change it exits immediately.

### DIFF
--- a/src/iai_mcp/socket_server.py
+++ b/src/iai_mcp/socket_server.py
@@ -77,6 +77,7 @@ class SocketServer:
         self.last_activity_ts: float = time.monotonic()
         self.active_connections: int = 0
         self.shutdown_event: asyncio.Event = asyncio.Event()
+        self._handler_tasks: set[asyncio.Task[Any]] = set()
         self._state = state
 
     async def handle(
@@ -85,6 +86,9 @@ class SocketServer:
         writer: asyncio.StreamWriter,
     ) -> None:
         self.active_connections += 1
+        handler_task = asyncio.current_task()
+        if handler_task is not None:
+            self._handler_tasks.add(handler_task)
         try:
             while not reader.at_eof():
                 line = await reader.readline()
@@ -255,12 +259,30 @@ class SocketServer:
             pass
         finally:
             self.active_connections -= 1
+            if handler_task is not None:
+                self._handler_tasks.discard(handler_task)
             try:
                 writer.close()
                 await writer.wait_closed()
             except (OSError, ConnectionError):  # noqa: BLE001 -- cleanup is best-effort
                 pass
 
+
+    async def _drain_handler_tasks(self) -> None:
+        """Cancel live connection handlers and wait for them to unwind.
+
+        ``Server.wait_closed()`` returns only once every active connection has
+        been dropped (CPython 3.12.1+; before that it returned immediately even
+        with connections open). A handler parked on ``reader.readline()`` for an
+        idle client never drops on its own, so without this the shutdown blocks
+        until the supervisor's stop timeout and the daemon dies by SIGKILL.
+        """
+        handlers = [t for t in self._handler_tasks if not t.done()]
+        if not handlers:
+            return
+        for task in handlers:
+            task.cancel()
+        await asyncio.gather(*handlers, return_exceptions=True)
 
     async def serve(self, socket_path: Path | None = None) -> None:
         if socket_path is None:
@@ -279,9 +301,11 @@ class SocketServer:
                 self.handle, socket_path, limit=_line_limit,
             )
             try:
-                async with server:
+                try:
                     await self.shutdown_event.wait()
+                finally:
                     server.close()
+                    await self._drain_handler_tasks()
                     await server.wait_closed()
             finally:
                 shutdown_ipc()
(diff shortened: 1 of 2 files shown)


CodeAbra:
great catch, and thanks for the depth — the 3.12.1 wait_closed change and the cancellation landing before the `async with` body even runs is exactly it. the repro and task dump made it easy to confirm.

we've landed a fix based on your analysis, with one tweak: we don't cancel a handler that's mid-request, so a stop can't cut off an in-flight write. same root cause, same idea.

so we'll close this unmerged, but you're credited in the changelog and it ships next release. really appreciate this and #158.

farislmn:
Cool! Can't wait for the update!

CodeAbra:
Fixed in v3.1.0 — daemon shutdown no longer hangs on idle socket handlers. Thanks @farislmn for the precise diagnosis and the regression test.

